PR #36, opened 2015-01-09
Sparse CSR matrix

### LINKED ISSUE #34: SparseCSR class in Python
* Just a data structure containing:
    * val
    * channel (col_ind)
    * spike (row_ptr)
* `to_dense()`
* `from_dense`

rossant:
Just implements a container structure, some consistency checking, and tests.

To be tackled in future PRs as a function of our needs:
- selection of rows/columns
- mathematical operations
- conversion dense <-> sparse

Closes #34.


rossant:
Ready for review/merge @nippoo.

phy/io/sparse.py
```diff
@@ -0,0 +1,92 @@
...
+    return True
+
+
+class SparseCSR(object):
+    """Sparse CSR matrix data structure."""
+    def __init__(self, shape=None, data=None, channels=None, spikes_ptr=None):
+        # Ensure the arguments are all arrays.
+        data = np.asarray(data)
+        channels = np.asarray(channels)
+        spikes_ptr = np.asarray(spikes_ptr)
```
nippoo: why spikes_ptr? not indices, indptr, data? this will be used for features & waveforms and maybe more...

rossant: Because I think we'll always use this structure with spikes on the first axis and channels on the second axis. That's the case for features and waveforms at least.

PR #53, opened 2015-01-27
Add CCG computation

rossant:
- Add CCGs and tests
- Some refactoring

### DIFF
--- a/phy/electrode/mea.py
+++ b/phy/electrode/mea.py
@@ -54,3 +54,14 @@ class MEA(object):
     @adjacency.setter
     def adjacency(self, value):
         self._adjacency = value
+
+
+#------------------------------------------------------------------------------
+# Common probes
+#------------------------------------------------------------------------------
+
+def staggered_positions(n_channels):
+    """Generate channel positions for a staggered probe."""
+    i = np.arange(n_channels - 1)
+    x, y = (-1) ** i * (5 + i), 10 * (i + 1)
+    return np.flipud(np.r_[np.zeros((1, 2)), np.c_[x, y]])
--- a/phy/stats/ccg.py
+++ b/phy/stats/ccg.py
@@ -7,7 +7,6 @@
 #------------------------------------------------------------------------------
 
 import numpy as np
-from numpy.lib.stride_tricks import as_strided
 
 from ..ext import six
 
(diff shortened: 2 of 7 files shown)


rossant:
Should be ready for review/merge.

phy/electrode/mea.py
```diff
@@ -54,3 +54,14 @@ def adjacency(self):
     @adjacency.setter
     def adjacency(self, value):
         self._adjacency = value
+
+
+#------------------------------------------------------------------------------
+# Common probes
+#------------------------------------------------------------------------------
+
+def staggered_positions(n_channels):
```
nippoo: why specifically this and not tetrodes...?!

phy/utils/tests/test_array.py
```diff
@@ -40,6 +40,13 @@ def test_normalize():
     assert positions_n.max() <= 1
 
 
+def test_index_of():
+    """Test _index_of."""
+    arr = [36, 42, 42, 36, 36, 2, 42]
+    lookup = _unique(arr)
+    _index_of(arr, lookup)
```
nippoo: you're just testing that... it doesn't return an error?

phy/electrode/mea.py
rossant: we'll add many more patterns later, I just needed one for testing waveform view

phy/utils/tests/test_array.py
rossant: oops... gonna fix that right now

rossant: done, thx!

phy/stats/ccg.py
```diff
@@ -0,0 +1,133 @@
...
+from ..ext import six
+from ..utils.array import _index_of, _unique
+
+
+#------------------------------------------------------------------------------
+# Cross-correlograms
+#------------------------------------------------------------------------------
+
+def _increment(arr, indices):
+    """Increment some indices in a 1D vector of non-negative integers.
```
nippoo: assert that the indices aren't out of range? i.e. that indices doesn't contain something more than len(arr)?

rossant: We could do that but I fear that it would badly affect performance. This is called during a performance-critical loop during the computations of CCGs...

nippoo: hmmm fair enough. I wonder whether we could create ```d_assert```, ```d_ae``` functions or something (i.e. debug_assert) - if they called a function which checks whether ```DEBUG == True``` then would they get optimised away? Or would there still be an overhead due to context switch or something? In C the compiler would just optimise that away altogether if DEBUG is set to True...

rossant: Yes I think that's something we can do in the near future. There would be no overhead with such an `if`, here the overhead would probably be due to an array operation like `max()` on a big array.

PR #55, opened 2015-01-28
Session class

rossant:
Add a `Session` class providing user-level clustering actions.

This class makes sure all objects and views are synchronized, and it is responsible for updating everything following user actions.

This PR just sets up the structure. Things that remain to be done later include:
- Update the statistics
- Try integrating actual views
- (optional) Update selectors


rossant:
Rebased and solved conflicts.

PR #100, opened 2015-02-09
Refactor Session and cluster metadata

### LINKED ISSUE #97: Refactor Session
* A base class that implements a simple event system to register update callbacks after load, select, and cluster, and also a global undo stack.
* A default sorting class with selector, model, cluster info, etc.
* A user API to add custom functionality to the session controller (for example, add a new cluster statistics).
* `Selector.update(up)` should be defined in the default session; there should be a mechanism where the object that instanciates `Selector` can also implement an update function:

    ```python
    @session.register(Selector)
    def on_cluster(selector, up):
        # Select the new merged cluster after a merge.
        selector.select_clusters(up['added'])
    ```

rossant:
- BaseClusterInfo is a generic class holding cluster-dependent data.
- ClusterMetadata derives from it.
- ClusterStats too. It acts like a cache; the defaultdict factory function acts as a function computing a cluster statistic, only called if the cluster is not in cache.
- Closes #97.


rossant:
ready for merge @nippoo

PR #103, opened 2015-02-11
A few updates in waveform view

rossant:
* [x] Horizontal box scaling
* [x] Aspect ratio is now kept in normalization of channel positions


nippoo:
Yeah. I'm not really sure when we'll want to generate a 'random color' though. No harm in having the function: *but* it would be more helpful to pre-generate or algorithmically generate guaranteed contrasting sets, maybe. My vote would be to limit multi-selections to, say, 20 clusters and have a set of 20 guaranteed distinct colours...

rossant:
@nippoo yes, random color is a nice-to-have feature for quick tests, but that's not necessarily what we'll end up using eventually.

PR #116, opened 2015-02-24
Cluster store

### LINKED ISSUE #58: Efficient data structures for the features
Benchmarks need to be done in order to find efficient on-disk formats for the features.

* Features are used for:
    * Feature View (a subset of the spikes, two features x and y)
    * Split action (find all spikes which features x and y within a given polygon)
    * Similarity matrix (a subset of the spikes, but all feature columns)
 
Example size (high estimate): a `(n_spikes, n_features)` numerical matrix with:

* `n_spikes = 100,000,000`
* `n_features = 10,000`
* about 20 non-null values per spike (sparse array)
* float32 data type
* total size (sparse): ~10 GB

Access patterns:

1. View: arbitrary subset of <10,000 of rows, 2 arbitrary columns x and y.
2. Split: arbitrary subset of several 10,000s of rows, 2 arbitrary columns x and y.
3. Matrix: regular subset of ~10,000 rows (strided selection), all columns.

Possibilities:

* HDF5 (dense, sparse csr, something else)
* sqlite
* flat binary

Notes:

* Possibility to duplicate the data on disk using different structures for different access patterns.
* Possibility to cache up to X GB of data, with X being a user option (1 by default?), the larger X, the better the performance.
* We can consider SSDs exclusively for benchmarks.

### LINKED ISSUE #70: Make clusters and their spikes deletable
For performance and memory considerations, it may be needed to delete noisy spikes to save memory.

### LINKED ISSUE #99: Data structure to store a list of spikes per cluster
* We keep `spike_clusters` for the undo stack.
* We add another structure `cluster ==> list of spikes` that is updated at every cluster assignement. This allows for fast selection of spikes.

### LINKED ISSUE #114: ClusterStore
In-memory structure storing per-cluster information that only depends on the recording data (i.e. not user-dependent information like cluster group, which comes with its own undo stack)

Closes #58, closes #70, closes #99.

* Some of this information may be memory-mapped (like the list of all features and masks per cluster), using an on-disk data store

* Every piece of information can be in a MemoryStore or a DiskStore. This is transparent in the external API. Every field choses how it's going to be stored.

* Modular architecture: it is easy to add/modify the kind of information stored in the structure

* Every cluster is identified by an `id`, which is a non-negative integer

* Fundamental hypothesis: the information of cluster `<id>` cannot change during a clustering session. New clusters can be created, and clusters can be deleted, but no cluster is modified. Merging/splitting => delete/new clusters.

* TODO: currently, `Clustering.assign()` *can* result in clusters being modified without their IDs being changed (for example, split). This need to change!

* This structure is completely decoupled from the file format. There are import/export functions from/to the file format.

* There is an `update(up)` method that is called when clustering assignements change (merge/split/assign). The structure should update accordingly.

* Initial list of fields in the structure; for each cluster:
  * n_spikes (memory)
  * spike_ids (memory)
  * features (disk)
  * masks (disk)
  * waveforms (raw data on disk)
  * mean_masks (memory)
  * mean_waveforms (memory)
  * quality (memory)  # find a more specific name
  * n_unmasked_channels (memory)
  * mean_probe_position (memory)

* Rename `spike_label` and `cluster_label` to `spike_id` and `cluster_id`

* There are several events:
  * `import`: when data is initially imported and all cached data is cleared
  * `export`: when internal data is exported to a public file
  * `update`: when clusters change

* There are several methods:
  * `_store(cluster, ...)`
  * `_load(cluster, ...)`
  * `delete(clusters)`

* Every field can subscribe to any event:

  ```python
  @cluster_store.connect
  def on_import(model, cluster):
      """Called when data is imported, for every cluster."""
      # get my_features for the given cluster from the model
      self._store(features=my_features, cluster=cluster,
                  storage='disk')

  @cluster_store.connect
  def on_import(model):
      """Called when data is imported, just once."""
      self._store(..., storage='memory')

  @cluster_store.connect
  def on_update(up):
      if up['description'] == 'merge':
          self._store(...)
          self._delete(...)

  @cluster_store.connect
  def waveforms(cluster):
    self._load(name='waveforms', cluster=cluster)

  @cluster_store.connect
  def mean_masks(cluster):
    self._load(name='mean_masks', cluster=cluster)
  ```

* For disk storage, we need a private class handling file management (one file per cluster):

  ```python
  # Create the folder if it doesn't exist
  ds = DiskStore(some_folder)

  ds.store(features=my_features, cluster=cluster)
  ds.load(name='features', cluster=cluster)
  ds.delete(clusters)

  ```

* Idem for MemoryStore (very simple: just wrapping a dict).

* Function to extract a subarray from a sparse CSR into a SCSC (simplified SCSC).

* TODO: progress bar API (when initially importing a dataset)

ping @nippoo @ctaf42

rossant:
This PR brings significant refactoring changes and new features. The clustering logic has been strengthened, and there is now a ClusterStore that holds per-cluster information in memory and on disk.
- [x] MemoryStore
- [x] DiskStore
- [x] Generic event emitter, Session derives from it
- [x] BaseClusterStore with event system
- [x] _all_spikes_in_clusters()
- [x] Clustering now has an in-sync spikes_per_cluster structure
- [x] ProgressReporter
- [x] Rename `label` to `id` for clusters and spikes
- [x] Update cluster assign such that modified clusters get a new id
- [x] Transform interface session into a class
- [x] Add descendants adjacency list in UpdateInfo: useful to track the genealogy of clusters
- [x] More clustering tests
- [x] Session ==> BaseSession and Interface ==> KwikSession. Merge everything into the same module.
- [x] Actual ClusterStore
- [x] Refactor loading/updating per-cluster data in the session

Closes #58. Closes #70. Closes #99. Closes #114.


rossant:
OK @nippoo this is ready to review/merge! There are still a few things to do but I'll do them later (#117). This PR is huge enough.

PR #134, opened 2015-04-01
Feature view

### LINKED ISSUE #30: Basic FeatureView
* Just a scatter plot of selected spikes
* Subset of all spikes from a list of given clusters
* Point colors as a function of the cluster
* Refactor WaveformVisual in a BaseVisual with baking mechanism

rossant:
## Description

This view displays a grid of `n_rows x n_rows` plots (origin at the top left). A column or a row is defined by a *dimension* which can be either:

* `(channel_id, feature)`: this specifies a given value for the spike
* `'time'`: this is the time of the spike

## Inputs

* `features`: an `(n_spikes, n_channels, n_features)` array (dense for now, sparse later)
* `masks`: an `(n_spikes, n_channels)` array
* `spike_ids`: the identifiers of the spikes
* `channel_ids`: the identifiers of the channels
* `spike_times`: the spike times of the selected spikes (`float64` in seconds)
* `spike_clusters`: the cluster index of every spike
* `cluster_colors`: the cluster of every color
* `dimensions`: a list of dimensions

## Notes

Closes #30.

## Screenshot

image


rossant:
This [VisPy PR](https://github.com/vispy/vispy/pull/854) needs to be merged before travis is happy here.

rossant:
@nippoo this is ready for review/merge
@shabnamkadir this one is for you! (look at the screenshot)

phy/plot/features.py
```diff
@@ -0,0 +1,242 @@
...
+        boxes = np.hstack(boxes)
+
+        assert positions.shape == (n_points, 2)
+        assert masks.shape == (n_points,)
+        assert boxes.shape == (n_points,)
+
+        self.program['a_position'] = positions.copy()
+        self.program['a_mask'] = masks
+        self.program['a_box'] = boxes
+        self.program['u_size'] = 5.  # TODO: config
```
nippoo: This needs to be scaled up for HiDPI support? This will appear half as small on my screen, unless I'm missing something.

rossant: Maybe.. how should I do? Can you put a screenshot? 
In any case you can press +/- to change the size!

nippoo: Well, VisPy does it using the `canvas.physical_size` property returned by the backend or the `pixel_scale` property (if using the AppCanvas). SceneCanvas does this with the transforms? Basically would be `self.program['u_size'] = 5 * canvas.pixel_scale`...

rossant: hmm the problem is that I don't have access to the canvas from within the visual... I prefer to keep it as it is now

nippoo: hmm okay - take a look at @Eric89GXL's code on this: https://github.com/vispy/vispy/pull/826/files#diff-fbc370cd9b515997708a2d0ac2d63b5cR212

I don't know if that's directly applicable but...

Eric89GXL: the transform system is usually used to get this information, which is what those referenced lines should do... not sure if it would work for this code or not

phy/plot/_vispy_utils.py
```diff
@@ -12,175 +12,418 @@
...
+                getattr(self, name)()
+        self._to_bake = []
+        return n_bake > 0
+
+    def draw(self):
+        """Draw the waveforms."""
+        # Bake what needs to be baked at this point.
+        self._bake()
+        if self.n_spikes is not None and self.n_spikes > 0:
+            self.program.draw(self._gl_draw_mode)
```
rossant: @Eric89GXL the visual's draw method is here

phy/plot/_vispy_utils.py
```diff
@@ -12,175 +12,418 @@
...
+        self._pz = PanZoom()
+        self._pz.add(self.visual.program)
+        self._pz.attach(self)
 
     def on_draw(self, event):
-        gloo.clear()
-        for visual in self.visuals:
-            visual.draw(self._tr)
+        gloo.clear('black')
+        self.visual.draw()
```
rossant: Should I pass the transform stuff here? Then maybe I can get access to the dpi information? @Eric89GXL

Eric89GXL: Usually you pass a TransformSystem that has information about the different coordinate frames. If you're using the SceneCanvas you get it automatically, if not then you construct it yourself.

Check out the `TextVisual` example, which doesn't use SceneCanvas at all, to see how it can be done:

https://github.com/vispy/vispy/blob/master/examples/basics/visuals/text_visual.py

This example is designed to draw text at 48 point size (initially), i.e. 48/72 of an inch, on every screen. Then the text visual is set up to use the TransformSystem to get the right number of physical pixels to draw to get it to be that physical size:

https://github.com/vispy/vispy/blob/master/vispy/visuals/text/text.py#L464

Note that the `LineVisual` also uses the transform system to be able to scale up the `glLineWidth` call if necessary:

https://github.com/vispy/vispy/blob/master/vispy/visuals/line/line.py#L354

rossant: Thanks, it works!

rossant:
@nippoo I won't implement the dpi stuff in this PR, so feel free to merge when you can!

PR #160, opened 2015-04-07
Updates and bug fixes

rossant:
* [x] View models handle how data is passed to views.
* [x] Fixed CCGs computation and display.
* [x] Improved cluster store.
* [x] Smart projections in feature view, as a function of the best channels for the selected clusters.
* [x] Simple qualitative browser colormap for selected clusters.
* [x] Rename `_non_empty` to `_empty` in views.
* [x] Rename `views.py` to `view_model.py`.
* [x] Refactor utility in `cluster/manual/`.
* [x] Refactor `store_from_model`.
* [x] Get ride of `Store`.
* [x] Optimize caching of features and masks in cluster store.
* [x] Port over #161.

## Issues closed

* Closes #146.
* Closes #147.
* Closes #152.
* Closes #157.
* Closes #163.

PR #161, opened 2015-04-07
session: support close

cgestes:


rossant:
thanks. I ported the changes to #160, so I'm going to close thise

Issue #197, opened 2015-04-14
Improve consistency check of the cluster store


rossant:
see https://github.com/kwikteam/phy/issues/196#issuecomment-92718725

nippoo:
YAGNI

PR #202, opened 2015-04-14
Updates in cluster store and views

### LINKED ISSUE #159: Cluster store manager
* Display the total size of all cluster stores
* Display the size of each dataset's cluster store, with per-cluster detail
* Clear cluster stores of some datasets
* Clean cluster store of some datasets (delete old clusters)
* Regenerate the cluster store of some datasets

### LINKED ISSUE #176: store.main_channels() should sort the channels by decreasing order of importance

### LINKED ISSUE #191: Better interactivity in the views
Improve current implementation of pan and zoom.

### LINKED ISSUE #193: Rename model.filename to model.kwik_path

### LINKED ISSUE #194: Choose between clusters or cluster_ids
Currently, both are used interchangeably. We should stick to one.

@nippoo for example, in the `Clustering` class, the list of current cluster identifiers is `cluster_ids`. Should we choose `clusters` or `cluster_ids` in all such instances?

### LINKED ISSUE #196: More robust cluster store generation
When loading a dataset, check if the cluster store exists. If not, create it. Otherwise, check the file size of every file. If all's good, do not regenerate it. Otherwise, if there's an inconsistency, display a warning, and recreate the corrupted or missing files.

The user should be warned that changing the spike clusters manually necessitates to force the regeneration of the cluster store.

### LINKED ISSUE #199: Complete docstring of KwikModel
So that `model?` shows the list of properties in the IPython notebook.

### LINKED ISSUE #200: Support multiple clusterings in Session

### LINKED ISSUE #205: Save size and position of VisPy canvases when using desktop backends

rossant:
* Closes #159 
* Closes #176
* Closes #191
* Closes #193
* Closes #194
* Closes #196
* Closes #199
* Closes #200
* Closes #205


rossant:
@nippoo this is ready for review

rossant:
Dealing with multiple clusterings should "just work" now @shabnamkadir , let me test it properly in a notebook first;

phy/cluster/manual/session.py
```diff
@@ -201,18 +188,36 @@ def _store_cluster(self,
                               append=True,
                               )
 
-    def store_all_clusters(self, spikes_per_cluster):
+    def is_consistent(self, cluster, spikes):
+        """Return whether the data of a cluster is consistent."""
```
nippoo: In fairness this is probably more "Return whether the filesize of a cluster store file is correct" - you're not checking the data at all, even less for consistency!

nippoo: You could MD5 both the cluster and the store object, but I imagine that'd take longer!

rossant: > In fairness this is probably more "Return whether the filesize of a cluster store file is correct" - you're not checking the data at all, even less for consistency!

Good point I'll change it

rossant: > You could MD5 both the cluster and the store object, but I imagine that'd take longer!

yes see #197

rossant: > In fairness this is probably more "Return whether the filesize of a cluster store file is correct" - you're not checking the data at all, even less for consistency!

on the other hand, I feel that just checking the file size is sufficient in 99.9999% of the cases. Why? Because:

* We check for consistency of *all* files: if a single file doesn't have the right size, the cluster store is declared inconsistent
* Cluster ids are unique: when you change a cluster, its identifier changes. So if you have two cluster stores with the same number of files, the same ids, and the same number of bytes in every file, I can hardly see how the clusterings can be different...

nippoo: I don't disagree. But it's still misleading to imply it's a data consistency check! I reckon MD5 hashing each cluster would probably be more hassle / time than it's worth.

phy/cluster/manual/clustering.py
```diff
@@ -113,56 +165,83 @@ def __init__(self, spike_clusters):
...
+
+        All changes are lost.
+
+        """
+        self._undo_stack.clear()
         self._spike_clusters = self._spike_clusters_base
         self._update_all_spikes_per_cluster()
 
     @property
     def spike_clusters(self):
```
nippoo: it's rare that we need this, right?

rossant: wdym?

rossant:
docstrings added! cc @nippoo @shabnamkadir

rossant:
This is done, ready to be merged on my end

phy/cluster/manual/clustering.py
nippoo: well: under what situations do we need to access one long contiguous array with spike_clusters? it's FAR more often that you want to access the individual spikes times inside one cluster, right?

rossant: Not really. For example, the undo stack is fundamentally based on this array. Also, you need this array when you want to compute all pairwise CCGs in a set of neurons. Finally, this is what is stored in the file format.

rossant: That being said, `session.clustering.spikes_per_cluster` contains what you want, and is always synchronized with `spike_clusters`.

PR #224, opened 2015-04-26
Trace view

### LINKED ISSUE #16: Trace viewer
[Possible starting point](https://github.com/kwikteam/experimental/tree/master/plot). Based on Vispy.

### Features

* Simple paging system.
* Load the entire page into GPU memory, no dynamic undersampling (first approach).
* Load and show the previous and next pages.
* Pan & zoom.
* Change channel scaling uniformly.
* Optional automatic page scrolling with a timer.

### Inputs

* NumPy array (or memmap array) of size `(nchannels, nsamples)`
* h5py dataset
* [Optional] spike trains (spike times, neuron indices, masks) to show the spikes within the traces

### Options

* Color of the channels
* Page size

### LINKED ISSUE #175: Minor display bug in feature view
The background of some disc markers is sometimes not transparent. This may be due to transparency and depth testing. Check discard in marker fragment shader.

### LINKED ISSUE #177: Spike selection subsets should have the same number of spikes per cluster
I'll fix it this week.

### LINKED ISSUE #198: Option to only zoom in the subplot hovered by the mouse pointer
* [x] Independent panning/zooming in the subplots
* [x] alt+wheel = global zooming
* [x] anisotropic zooming with right click

Can use a uniform with the index of the currently-hovered subplot, and if statement for pan/zoom in GLSL.

### LINKED ISSUE #218: Bug with multiple canvases
Try using `canvas.gloo.func` instead of `gloo.func`

### LINKED ISSUE #223: Decouple spike and cluster selector
The spike selector should be the responsibility of the view model rather than the session.

### LINKED ISSUE #227: warn when correct files aren't present
Would be nice to be informed if necessary files aren't found (e.g. raw.kwd).

### LINKED ISSUE #228: Order of selected clusters and displayed clusters in waveform view should match
Supersedes #226.

rossant:
This PR implements the trace view. It also brings many other improvements and bug fixes.
- Trace view with user-settable interval and spikes with masks.
- Refactored selector: now, each view model is responsible for its selector instance. There is no more global selector instance in Session.
- A view no longer has a reference to the full spike_clusters array, just the spike clusters for the spikes shown in that view.
- Warning if a kwik file is not present.
- Improved pan zoom: in a grid view, pan zoom is now specific to each subplot (using a matrix of pan/zoom parameters with an OpenGL uniform array). There are a few global keyboard shortcuts.
- Refactored pan zoom is a dedicated private module.
- There is now a `PanZoomGrid` class that derives from `PanZoom`.
- Added pan constraints.
- Added special pan constraints for time subplots in feature view.
- Global zoom applies to a single axis in time subplots.
- Fixed display bug with feature markers.
- Take the cluster order into account in the views.

Issues closed:
- Closes #16 
- Closes #175
- Closes #177
- Closes #198 
- Closes #218
- Closes #223 
- Closes #227 
- Closes #228


rossant:
@nippoo this is ready for review/merge!

phy/cluster/manual/view_model.py
```diff
@@ -74,34 +98,70 @@ def store(self):
...
+
+    def _update_cluster_order(self, up):
+        """Update cluster order when a clustering action occurs."""
+        self._view.visual.cluster_order = _update_cluster_selection(
+            self._view.visual.cluster_order, up)
 
     def on_open(self):
-        """To be overriden."""
-        self._update_spike_clusters()
+        """May be overriden."""
```
nippoo: does this even run?!

rossant: Yes: this is not a comment, but a string. It's a bit like a no-op.

phy/plot/_panzoom.py
```diff
@@ -0,0 +1,794 @@
...
+
+import numpy as np
+
+# from vispy import gloo
+
+from ..utils.array import _as_array
+
+
+#------------------------------------------------------------------------------
+# PanZoom class
```
nippoo: Shouldn't this really be in vispy?!

rossant: Eventually, yes.

PR #234, opened 2015-05-01
Manual clustering GUI

### LINKED ISSUE #57: Remove unused imports
* Consider putting back flake8 [F401](http://flake8.readthedocs.org/en/2.2.3/warnings.html)
* Use [autoflake](https://github.com/myint/autoflake) once on the whole code base to remove unused imports.

### LINKED ISSUE #128: Clustering GUI
The goal of this issue is to implement (in phyui?) a clustering GUI based on PyQt and IPython notebook. The GUI wraps a `Session` instance which implements the logic. The logic (Session) and the GUI code are completely separated. You don't have to use the GUI to work with your data: you can just use the Session API for this.

The [`Session`](https://github.com/kwikteam/phy/blob/master/phy/cluster/manual/session.py) class implements the clustering logic:

* `session.open(file)` 
* `session.select(clusters)`
* `session.merge(clusters)`
* `session.undo()`
* etc.

There is no graphical/visualization part involved in `Session`.

The goal of the `GUI` class is to wrap a Session and implements an actual GUI, in HTML, Qt, or anything else.

Here is the base class:

```
class BaseGUI(object):
    def __init__(self, session):
        self._session = session

    @property
    def model(self):
        return self._session.model

    @property
    def session(self):
        return self._session

    def show(self):
        raise NotImplementedError()
```

Then we have a `ClusteringGUI(BaseGUI)` that implements the clustering GUI with PyQt for the main window and the VisPy canvases, and HTML/IPython/JS/D3 for the cluster view. Later on we'll have a `TemplateMatchingGUI(BaseGUI)` as well.

The GUI can be used like this:

```
gui = ClusteringGUI(session)
gui.show()  # display the QMainWindow with the default VisPy views and the ClusterView as well in the notebook (an error is raised if we're not in the notebook)

gui.show_cluster_view()  # just display the cluster view in the current notebook cell
gui.add_waveform_view()  # add a waveform view in the QMainWindow (there can be several views of the same type in the GUI)
gui.add_feature_view()  # idem
gui.add_correlograms_view()  # idem

gui.feature_views()  # return a list of Feature view instances

gui.session.open(file)  # this just opens a file using the wrapped Session instance
gui.session.merge(clusters)  # this just merges some clusters

gui.prompt_file()  # opens a FileDialog, the user selects a file, and the file is opened
gui.reset_views()  # reset the views to their default state
gui.save_views(file=None)  # save the current positions of the different views in a given file, or in a default location (e.g. ~/.phy/clustering_gui/qt_settings)

```

See some old code here (most of it won't be used, except some of the styles maybe):

* https://github.com/klusta-team/klustaviewa/blob/master/klustaviewa/gui/mainwindow.py
* https://github.com/klusta-team/klustaviewa/blob/master/klustaviewa/gui/dock.py
* https://github.com/klusta-team/klustaviewa/blob/master/klustaviewa/gui/styles.css

### LINKED ISSUE #222: Keyboard shortcuts for manual clustering

### LINKED ISSUE #229: Show views with fixed cluster selection
For example, `session.show_waveforms()` should show the selected clusters, while `session.show_waveforms(clusters)` should show the waveforms of the given clusters. The first view will change when the cluster selection changes, but the second won't.

### LINKED ISSUE #233: Revamped wizard
The wizard offers the functionality exposed in the GUI via keyboard shortcuts to quickly sort a dataset.

Use the wizard to:

* Go through all of your unsorted clusters
* Quickly find merge candidates
* Move clusters to quality groups

Features:

* start, pause, stop
* is_running
* current list of clusters (either best unsorted or closest match)
* next, previous, first, last in that list
* pin the current cluster: the current list becomes the list of closest matches
* unpin

Keyboard shortcuts (in phyui):

* space: if not is_running, start. Otherwise, next
* enter: pin the current cluster (only if there's no pinned cluster)
* backspace: unpin the current cluster (if there's a pinned cluster), or stop
* g: merge (if there are two selected clusters)
* alt+n/m/g: move the pinned cluster to noise/mua/good, unpin, start, and pin
* ctrl+n/m/g: move the closest match cluster to noise/mua/good, and next

rossant:
This PR brings a lightweight manual clustering graphical interface. The interface combines several VisPy views in a single window using docking widgets. There are also keyboard shortcuts to scroll through clusters and pairs of clusters, merge them, and move them according to their quality.

* Closes #57.
* Closes #128.
* Closes #222.
* Closes #229.
* Closes #233.

## Features

* Improved wizard with navigation, pinning, support for cluster groups.
* Generic lightweight PyQt main window that accepts any Qt widgets/VisPy views as docking widgets.
    * Actions with global keyboard shortcuts can be set.
    * Main window and docks geometry and state can be persisted.
* `session.show_gui()` opens a main window with all clustering views. This GUI directly exposes the wizard with keyboard shortcuts.
* Customizable keyboard shortcuts for wizard actions in the GUI.
* Wizard panel that shows the current best and current match, as well as progress bars, to given an idea of how much has been done so far.

## Screenshot of the GUI

image


rossant:
@Eric89GXL I'm seeing segmentation faults on travis when using PyQt -- just wondering if you had ever seen this or had an obvious idea of what may be going wrong at this stage? I'm not seeing this bug locally on Linux. Otherwise, how to debug this?

Eric89GXL:
The segfaults I've seen with `vispy` have been caused by sending GL commands after a context has been closed. Do you have multiple contexts at all?

rossant:
thanks @Eric89GXL. Finally I found the bug: I was creating a new Qt application in a new unit test, which would conflict with an existing Qt application created by the other unit tests based on VisPy.

rossant:
Ready for review/merge.

phy/cluster/manual/wizard.py
```diff
@@ -67,62 +99,395 @@ def set_quality_function(self, func):
...
+    def _delete(self, clusters):
+        for clu in clusters:
+            if clu in self._cluster_groups:
+                del self._cluster_groups[clu]
+            if clu in self._best_list:
+                self._best_list.remove(clu)
+            if clu in self._match_list:
+                self._match_list.remove(clu)
+
+    def _add(self, clusters, group):
```
nippoo: not sure why these should be private functions. Sure, they're rare operations, but still possible for the user to do, right?

phy/cluster/manual/wizard.py
```diff
@@ -67,62 +99,395 @@ def set_quality_function(self, func):
...
+                                  best_group=self._group(self.best),
+                                  match_group=self._group(self.match),
+                                  )
+
+
+#------------------------------------------------------------------------------
+# Wizard panel
+#------------------------------------------------------------------------------
+
+_PANEL_HTML = """
```
nippoo: this should probably be in a separate .html and .css file, like the vert/frag shaders.

PR #248, opened 2015-05-07
Prepare beta

### LINKED ISSUE #207: Use `` instead of '' in docstrings for code

### LINKED ISSUE #245: Move ClusterMetadata to phy.io
The undo/redo features may be implemented in a child class in cluster/manual.

### LINKED ISSUE #246: Utility to download test datasets

### LINKED ISSUE #256: Add ctrl/shift+wheel shortcut in waveform view
to scale waveforms

### LINKED ISSUE #258: Quick start guide
(in a notebook)

* Quick install
* Download test data with phy
* Start the session
* Display some info (n spikes, n clusters, length of recording) with the kwik model
* Start the wizard GUI

### LINKED ISSUE #259: model.describe() method
Shows the filename, path, list of channel groups, number if recordings, n spikes, list of clusterings, n clusters, rec length, probe name, number of dead/ignored channels, etc

### LINKED ISSUE #260: Support non-default cluster groups in wizard

rossant:
* API doc generation tool
* cluster metadata refactoring
* fix wizard undo stack
* export public objects
* more docstrings

## Closes

*  closes #207
* closes #245 
* closes #246 
* closes #256
* closes #258
* closes #259
* closes #260


phy/cluster/manual/_history.py
```diff
@@ -125,18 +131,18 @@ def __init__(self, process_ups=None):
         self.process_ups = process_ups
 
     def action(self, *controllers):
-        """Register one or several controllers for this action.
-
-        The order matters: the first controller does the first action, etc.
```
nippoo: the order no longer matters?

rossant: indeed. I thought I would be using this feature but it looks I won't... hopefully this bit can be simplified at some point..

phy/utils/dock.py
```diff
@@ -266,12 +273,27 @@ def _try_enable_ipython_qt():
...
+        return
+    try:
+        from IPython import get_ipython
+        ip = get_ipython()
+        ip.enable_gui('qt')
+        global _APP_RUNNING
+        _APP_RUNNING = True
+        info("Qt event loop activated.")
+    except:
+        warn("Qt event loop not activated.")
```
nippoo: should this be a warn, or fatal? otherwise it'll crash elsewhere no?

rossant: not necessarily. qt event loop integration is horribly hard to get right: it needs to work in 875877 cases...

PR #267, opened 2015-05-12
More features

### LINKED ISSUE #138: Spike selection in feature view
* Click on a spike to highlight it in all subplots (and in other widgets as well) (use picking technique with framebuffer)
* CTRL + click to select multiple spikes
* `reset()` function to unselect all spikes
* Polygon lasso

### LINKED ISSUE #168: Background spikes in feature view
Cache them in the cluster store when generating the store.

### LINKED ISSUE #178: Better automatic selection of the dimensions in the feature view
* feature amplitude
* user-settable

### LINKED ISSUE #182: Implement cluster superposition in the waveform view

### LINKED ISSUE #240: Consider caching the waveforms

### LINKED ISSUE #261: Channel selection from waveform view to feature view
1+click on a channel in waveform view changes the dimension in the first subplot of feature view

### LINKED ISSUE #262: API to make views communicate in a dock GUI
```
view.events.add(channel_click=Event)
```

```
@waveform_view.connect
def on_channel_click(e):
    for view in feature_views:
        view.dimensions[1] = e.channel_id
```

``` 
self.events.channel_click(channel_id=...)
```

### LINKED ISSUE #264: gui.select fails when there is only a single point in a cluster
Clusters with a single point may occur for pathological clusterings.

### LINKED ISSUE #269: Change bins in correlograms view
* Python API
* Keyboard shortcuts

rossant:
- Closes #138 
- Closes #168 
- Closes #178
- Closes #182
- Closes #240
- Closes #261 
- Closes #262 
- Closes #264
- Closes #269

### DIFF
--- a/phy/cluster/manual/session.py
+++ b/phy/cluster/manual/session.py
@@ -22,6 +22,7 @@ from ...utils.settings import SettingsManager, declare_namespace
 from ...io.kwik_model import KwikModel, cluster_group_id
 from ._history import GlobalHistory
 from .clustering import Clustering
+from .selector import Selector
 from ._utils import (ClusterMetadataUpdater,
                      _spikes_per_cluster,
                      _concatenate_per_cluster_arrays,
@@ -54,13 +55,13 @@ class FeatureMasks(StoreItem):
               ('mean_probe_position', 'memory'),
               ]
 
-    # Size of the chunk used when reading features and masks from the HDF5
-    # .kwx file.
-    chunk_size = None
-
     def __init__(self, *args, **kwargs):
         self._pr_disk = kwargs.pop('progress_reporter_disk')
         self._pr_memory = kwargs.pop('progress_reporter_memory')
+        # Size of the chunk used when reading features and masks from the HDF5
+        # .kwx file.
+        self.chunk_size = kwargs.pop('chunk_size')
+
         super(FeatureMasks, self).__init__(*args, **kwargs)
 
         self.n_features = self.model.n_features_per_channel
@@ -154,6 +155,7 @@ class FeatureMasks(StoreItem):
                               masks=m,
                               append=True,
                               )
+        # TODO: compute stats here?
 
     def is_consistent(self, cluster, spikes):
         """Return whether the filesizes of the two cluster store files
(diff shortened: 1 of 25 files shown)


phy/cluster/manual/session.py
```diff
@@ -341,6 +343,96 @@ def on_cluster(self, up=None):
...
+            debug("Loading {0:s}, cluster {1:d}...".format(self.name,
+                  cluster))
+            self.store_cluster(cluster,
+                               spikes=self._spikes_per_cluster[cluster],
+                               mode=mode,
+                               )
+            self._pr.value += 1
+        self._pr.set_complete()
+
+    def is_consistent(self, cluster, spikes):
```
nippoo: why does this need to be separate from the features/masks consistency check? or is it nicer to have separate consistency check functions per clusterstore type?

rossant: for features I know the number of spikes per cluster so  I use that for consistency check
for waveforms not all spikes are stored. so I need to store them explicitely, and I use that for consistency

phy/cluster/manual/tests/test_view_model.py
```diff
@@ -54,19 +72,20 @@ def _test_view_model(view_model_class, stop=True, **kwargs):
...
+        # Merge the clusters and update the view.
+        debug("Merging.")
+        up = clustering.merge(clusters)
+        vm.on_select(up.added)
+        show_test_run(vm.view, _N_FRAMES)
+
+        # Split some spikes and update the view.
+        debug("Splitting.")
+        spikes = spikes[::2]
+        up = clustering.assign(spikes, np.random.randint(low=0, high=5,
```
nippoo: we should avoid randomness in testing - they should be deterministic...

nippoo: perhaps pseudorandom with a fixed seed? in this case it's not the *end* of the world, but it's possible for the test to fail once and then re-run and it passes, which isn't ideal

rossant: * we can use a fixed seed
* it is nice to have some randomness in a few specific tests, but in this case we have to make sure that they are completely robust

PR #279, opened 2015-05-14
Refactor view and GUI creation in session

rossant:
* Refactor settings (considerably simpler)
* Refactor view models
* Refactor view creation in session
* Refactor GUI creation in session
* Fix bugs in wizard
* Test Kwik view models
* Bug fixes
* More tests

## Closes

* closes #242
* closes #243
* closes #265 
* closes #280

PR #280, opened 2015-05-15
Fix out-of-order HDF5 trace selection

rossant:
@nsteinme this should fix your bug


rossant:
Has been backported in #279, so this will be closed when that other PR will be merged

PR #288, opened 2015-05-21
Bug fixes

### LINKED ISSUE #283: Mock cluster store that directly access the model
Simplifies the code in view model by assuming that a cluster store is always available.

### LINKED ISSUE #285: Standalone cluster store
* Currently, the store is only usable from the manual clustering session. We should provide an API to create and access a custom cluster store for a given experiment. The user should be free to add the store items they want (features, waveforms, statistics) or create their own.
* Further, the current default cluster store path has no mention of "manual clustering". We should add it, because it depends on the manual clustering settings. Notably, the number of waveforms to show = the number of waveforms to store. Several cluster stores should be able to coexist for any given experiment.

### LINKED ISSUE #290: Implement a way to select features from the GUI

### LINKED ISSUE #291: Default features selection not useful
@nsteinme what do you mean by "useful"?

### LINKED ISSUE #292: Some waveform pan/zoom controls missing
@nsteinme which particular action[s] would you need which aren't currently implemented?

### LINKED ISSUE #293: Crash when no KWX

### LINKED ISSUE #295: Time plot in FeatureView should zoom out to slighly longer than the extents of the data to allow selection

### LINKED ISSUE #296: Unify access to masks, features, waveforms
There should be a single way to access these data for a given cluster, independently of:

* whether we have a mock or kwik model
* whether there is a store or not
* whether the kwx/kwd files are present or not

One possibility is to have a mock cluster store that is resilient to all these cases and returns default values if necessary (zeros or ones)

### LINKED ISSUE #297: number of features selected for each display for each cluster should be independent
Currently, by selecting 10k features total from both clusters, you get this weird effect where, when you compare one cluster with very many spikes to one with few, the one with few seems to just disappear (because indeed, a very small percentage of the 10k will belong to that cluster). It would be much better if it were 10k *per cluster*.

### LINKED ISSUE #298: autocorrelogram and crosscorrelogram should be computed for much more than 10k points (all, if possible)
It's again because if you have one unit with few points relative to the other, you completely lose its ccg. Furthermore, the number of CCG refractory violations is a relevant measure and you're simply not displaying them by selecting only a subset of points. 

Could the CCGs be computed up front and cached? For ACGs can they be computed with much higher number of points included?

### LINKED ISSUE #299: double-click to select a feature panel should keep the same zoom characteristics
First thing - the double-click to select a feature panel feature is *awesome*! :) But there is an issue -

E.g. when I start with this:

image

and double click the panel in the first row, second column, this one:
image

I get this:

image

which can be manually re-zoomed to this, like I wanted:

image

### LINKED ISSUE #300: background points not always visible in feature view
E.g. this image from my last issue - there are, in fact, background points along the y-axis there, look really close!: 

image

Guess they need to somehow re-scale their opacity depending on density? Not sure if that makes sense.

### LINKED ISSUE #303: UpdateInfo __repr__()

rossant:
This PR brings some bug fixes, optimizations, and a significant refactoring of the cluster store. `StoreItems` now have more freedom to store and load data. There is a new `PerClusterData` that stores spike-dependent data and allows to make per-cluster selection.

* closes #283
* closes #285
* closes #290
* closes #291
* closes #292
* closes #293 
* closes #295 
* closes #296 
* closes #297 
* closes #298 
* closes #299
* closes #300
* closes #303

### DIFF
--- a/phy/__init__.py
+++ b/phy/__init__.py
@@ -12,14 +12,6 @@ containing up to thousands of channels.
 # Imports
 #------------------------------------------------------------------------------
 
-# Define a mock @profile decorator to avoid NameErrors when profiling.
-import sys
-if sys.version_info[0] == 3:
-    import builtins
-else:
-    import __builtin__ as builtins
-builtins.__dict__['profile'] = lambda func: func
-
 import os.path as op
 from pkg_resources import get_distribution, DistributionNotFound
 
--- a/phy/__init__.py
+++ b/phy/__init__.py
@@ -12,6 +12,14 @@ containing up to thousands of channels.
 # Imports
 #------------------------------------------------------------------------------
 
+# Define a mock @profile decorator to avoid NameErrors when profiling.
+import sys
+if sys.version_info[0] == 3:
+    import builtins
+else:
+    import __builtin__ as builtins
+builtins.__dict__['profile'] = lambda func: func
+
 import os.path as op
 from pkg_resources import get_distribution, DistributionNotFound
 
--- a/phy/cluster/manual/_utils.py
+++ b/phy/cluster/manual/_utils.py
@@ -34,7 +34,7 @@ def update_info(**kwargs):
         description='',  # information about the update: 'merge', 'assign',
                          # or 'metadata_<name>'
         history=None,  # None, 'undo', or 'redo'
-        spikes=[],  # all spikes affected by the update
+        spike_ids=[],  # all spikes affected by the update
         added=[],  # new clusters
         deleted=[],  # deleted clusters
         descendants=[],  # pairs of (old_cluster, new_cluster)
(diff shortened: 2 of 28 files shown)


rossant:
@nippoo this is ready for review, not for merge yet. I have a few tiny things to do, but you can review already...

rossant:

.travis.yml
```diff
@@ -28,17 +28,8 @@ install:
...
-  # Install VisPy master.
-  - pushd ~
-  - git clone https://github.com/vispy/vispy.git
-  - cd vispy
-  - python setup.py install
-  - popd
-  # Install phy.
-  - python setup.py install
+  - pip install coveralls responses python-coveralls pytest-cov
+  - pip install git+https://github.com/vispy/vispy.git
```
nippoo: hey this is neat!

PR #310, opened 2015-05-27
Fix channel selection bug in Python 2 and Windows

rossant:


rossant:
Ready to merge @nippoo this one should not take too long to review

PR #316, opened 2015-05-29
KK2 integration

### LINKED ISSUE #221: Wrap KK2

### LINKED ISSUE #241: Display the number of spikes in the ACG view

### LINKED ISSUE #270: Move wizard HTML in separate files

### LINKED ISSUE #274: Clearer error when kwik does not exist
More clear error needed for this.

I like bananas.

dir/phytest/phy/phy/io/kwik_model.pyc in open(self, kwik_path, channel_group, clustering)
    648         # Open the KWIK file.
    649         self._kwik = self._open_h5_if_exists('kwik')
--> 650         if not self._kwik.is_open():
    651             raise ValueError("File {0} failed to open.".format(kwik_path))
    652         self._check_kwik_version()

AttributeError: 'NoneType' object has no attribute 'is_open'

### LINKED ISSUE #311: Create sparsify function for KK2

### LINKED ISSUE #318: add stats to wizard view
Maybe I am getting too impatient about the cluster view, but would it be possible to add a list of any registered statistics to the wizard view (perhaps only if the statistic has just one element)?

So right below the cluster number and progress bar, it would just very simply say:
n_spikes = 10457
n_unmasked_channels = 23
estimated_FPrate_ISI = 0.03

(if I register a statistic that computes the estimated false positive rate from the ISI distribution, which is what I am imagining doing).

### LINKED ISSUE #319: Modular and extendable view creation
* New hierarchy of view models: `BaseViewModel` > `VispyViewModel` > `WaveformViewModel` etc.
* Child classes must implement `_view_name` and `_view_class`
* To create a view, need to provide:
  * the view model name or class
  * a dictionary of keyboard arguments for its constructor
* The view creator manages the creation of any type of view
* `on_select(clusters)` in view model, and `on_open()`, `on_close()`, `on_cluster()`
* `HTMLView` class that gets initialized with a function returning HTML string, and property with CSS styles
* Custom views can be added from `user_settings.py`

Fixes #318

rossant:
* [x] Create features and masks sparse structures for KK2
* [x] Make view models work without store (needed to visualize data with changing clustering)
* [x] HTML view models
* [x] Simplify handling of wizard panel in view creator
* [x] New stats view

## Closes

* Closes #221
* Closes #241
* Closes #270 
* Closes #274 
* Closes #311 
* Closes #318
* Closes #319

@thesamovar so I have written and tested a function that takes dense features and masks arrrays, and returns a `RawSparseData` structure. How should we proceed next? From a `KwikModel`, we should be able to do `run(model, **params)` (in phy) which would call `sparsify_features_masks(model.features, model.masks)`. Currently the launching funtion in KK2 expects CLI arguments, but I'd need Python arguments for the parameters.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -22,7 +22,7 @@ install:
   # NOTE: cython is only temporarily needed for building KK2.
   # Once we have KK2 binary builds on binstar we can remove this dependency.
   - conda install pip numpy matplotlib scipy h5py pyqt pyzmq ipython requests pytest pip flake8 coverage cython
-  - pip install --pre klustakwik2
+  - pip install klustakwik2
   - pip install six vispy coveralls responses python-coveralls pytest-cov
 before_script:
     - export DISPLAY=:99.0
--- a/.travis.yml
+++ b/.travis.yml
@@ -20,7 +20,8 @@ install:
   - conda create -q -n testenv python=$TRAVIS_PYTHON_VERSION
   - source activate testenv
   - conda install pip numpy matplotlib scipy h5py pyqt pyzmq ipython requests pytest pip flake8 coverage
-  - pip install six klustakwik2 vispy coveralls responses python-coveralls pytest-cov
+  - pip install --pre klustakwik2
+  - pip install six vispy coveralls responses python-coveralls pytest-cov
 before_script:
     - export DISPLAY=:99.0
     - /sbin/start-stop-daemon --start --quiet --pidfile /tmp/custom_xvfb_99.pid --make-pidfile --background --exec /usr/bin/Xvfb -- :99 -screen 0 1400x900x24 -ac +extension GLX +render
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,11 +5,6 @@ python:
   - "3.3"
   - "3.4"
 install:
-  # - sudo apt-get update
-  # You may want to periodically update this, although the conda update
-  # conda line below will keep everything up-to-date.  We do this
-  # conditionally because it saves us some downloading if the version is
-  # the same.
   - if [[ "$TRAVIS_PYTHON_VERSION" == "2.7" ]]; then
       wget http://repo.continuum.io/miniconda/Miniconda-latest-Linux-x86_64.sh -O miniconda.sh;
     else
@@ -24,12 +19,8 @@ install:
   # Create the environment.
   - conda create -q -n testenv python=$TRAVIS_PYTHON_VERSION
   - source activate testenv
-  # Optional dependencies.
-  - conda install pip numpy matplotlib scipy h5py pyqt pyzmq ipython requests
-  # Testing dependencies.
-  - conda install pytest pip flake8 coverage
-  - pip install coveralls responses python-coveralls pytest-cov
-  - pip install git+https://github.com/vispy/vispy.git
+  - conda install pip numpy matplotlib scipy h5py pyqt pyzmq ipython requests pytest pip flake8 coverage
+  - pip install six klustakwik2 vispy coveralls responses python-coveralls pytest-cov
 before_script:
     - export DISPLAY=:99.0
     - /sbin/start-stop-daemon --start --quiet --pidfile /tmp/custom_xvfb_99.pid --make-pidfile --background --exec /usr/bin/Xvfb -- :99 -screen 0 1400x900x24 -ac +extension GLX +render
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,7 +19,9 @@ install:
   # Create the environment.
   - conda create -q -n testenv python=$TRAVIS_PYTHON_VERSION
   - source activate testenv
-  - conda install pip numpy matplotlib scipy h5py pyqt pyzmq ipython requests pytest pip flake8 coverage
+  # NOTE: cython is only temporarily needed for building KK2.
+  # Once we have KK2 binary builds on binstar we can remove this dependency.
+  - conda install pip numpy matplotlib scipy h5py pyqt pyzmq ipython requests pytest pip flake8 coverage cython
   - pip install --pre klustakwik2
   - pip install six vispy coveralls responses python-coveralls pytest-cov
 before_script:
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -3,5 +3,6 @@ include README.md
 
 recursive-include tests *
 recursive-include phy/plot/glsl *.vert *.frag *.glsl
+recursive-include phy/cluster/manual/static *.html *.css
 recursive-exclude * __pycache__
 recursive-exclude * *.py[co]
--- a/phy/__init__.py
+++ b/phy/__init__.py
@@ -26,7 +26,7 @@ from .utils.dock import enable_qt, qt_app
 
 __author__ = 'Kwik team'
 __email__ = 'cyrille.rossant at gmail.com'
-__version__ = '0.1.0'
+__version__ = '0.2.0.dev0'
 
 
 __all__ = ['debug', 'set_level']
--- a/phy/cluster/__init__.py
+++ b/phy/cluster/__init__.py
@@ -2,3 +2,5 @@
 # flake8: noqa
 
 """Automatic and manual clustering facilities."""
+
+from .launcher import run
--- a/phy/cluster/launcher.py
+++ b/phy/cluster/launcher.py
@@ -1,0 +1,45 @@
+# -*- coding: utf-8 -*-
+
+"""Automatic clustering launcher."""
+
+#------------------------------------------------------------------------------
+# Imports
+#------------------------------------------------------------------------------
+
+from ..utils.array import PartialArray
+from ..io.kwik.sparse_kk2 import sparsify_features_masks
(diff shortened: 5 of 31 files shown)


thesamovar:
```
num_starting_clusters = params.pop('num_starting_clusters')
data = ... # get RawSparseData object
data = data.to_sparse_data()
kk = KK(data, **params)
kk.cluster_mask_starts(num_starting_clusters)
clusters = kk.clusters
```

There are other things you might want to do, like create a log file, or get the current clusters as callbacks (events). For examples of that, take a look at https://github.com/kwikteam/klustakwik2/blob/master/dev/kk2_with_monitoring.ipynb and https://github.com/kwikteam/klustakwik2/blob/master/dev/testing.py for the moment. If there is something else you want to do, just let me know.

rossant:
KK2 launcher done
@thesamovar can you try it?

```python
from phy.cluster.launcher import run
from phy.io.kwik import KwikModel

model = KwikModel(filename)
spike_clusters = run(model, num_starting_clusters=10, **params)
```

thesamovar:
I used ``test_hybrid_120sec`` and got this:

```
---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
<ipython-input-1-b3802741f923> in <module>()
      2 from phy.io.kwik import KwikModel
      3 
----> 4 model = KwikModel('test_hybrid_120sec')
      5 spike_clusters = run(model, num_starting_clusters=100, **params)

c:\programming\phy\phy\io\kwik\model.pyc in __init__(self, kwik_path, channel_group, clustering)
    294         self.open(kwik_path,
    295                   channel_group=channel_group,
--> 296                   clustering=clustering)
    297 
    298     # Internal properties and methods

c:\programming\phy\phy\io\kwik\model.pyc in open(self, kwik_path, channel_group, clustering)
    586         # Open the KWIK file.
    587         self._kwik = self._open_h5_if_exists('kwik')
--> 588         if not self._kwik.is_open():
    589             raise ValueError("File {0} failed to open.".format(kwik_path))
    590         self._check_kwik_version()

AttributeError: 'NoneType' object has no attribute 'is_open'
```

rossant:
add ".kwik"

2015-05-29 21:55 GMT+02:00 Dan Goodman <notifications@github.com>:

thesamovar:
Ah, I thought it was the directory name I should pass for some reason. OK, it seems to be running now (although of course I have no way of knowing if the sparse data structure it is working on is correct). One observation: when running from IPython you get all the log output which probably is going to kill us for memory. Also, my Firefox was running at 100% CPU (well, 25% because it is quad core and ff only uses one core).

thesamovar:
Also that error message is not very informative. ;)

thesamovar:
What is ``ipp_view``? Is that for the parallel stuff? I haven't implemented that yet.

rossant:
> One observation: when running from IPython you get all the log output which probably is going to kill us for memory. Also, my Firefox was running at 100% CPU (well, 25% because it is quad core and ff only uses one core).

maybe we can disable this output by default? (i.e. put that in DEBUG mode while the default is INFO?)

> Also that error message is not very informative. ;)

i knoww

> What is ipp_view? Is that for the parallel stuff? I haven't implemented that yet.

yes. even if it is not implemented you can add the argument and not use it, at least it won't crash if i pass the object to KK

thesamovar:
I don't think it should be moved from info to debug, because (a) it is very useful for monitoring what is going on while KK runs, (b) there is a HUGE amount of debug output (that is important for understanding what is going when something goes wrong) so those messages would be lost in all that. One thing you can do to reduce it a bit is this (which is included in the script by default):

```
log_suppress_hierarchy('klustakwik', inclusive=False)
```

With that change, you'll get a few messages at the beginning and then just one or two per iteration. That's still a lot for the browser though. I think the best way to deal with it is when we have a monitoring GUI for while KK runs, the log output should be in a scrolling window. For the moment, a quick hack would be to have an ``IPython.display.clear_output()`` every ten iterations or so? (Using a callback to do this.)

For the ``ipp_view`` can you just not pass this parameter to KK until it exists? Seems more sensible to do it that way than to have a fake parameter that is being ignored? I can certainly make the change in KK2 but it seems a bit strange to do it that way.

rossant:
How can I clear the log output exactly from [here](https://github.com/kwikteam/phy/pull/316/files#diff-27dc998be171a48b082b9a9ce5bd7feaR25) ?

Also, is the log saved in a file? If so, there might not be a need to show everything in the notebook.

Finally, by appending `\r` to a log line you can overwrite to the same line, which might be useful if you just want to show the very last lines.

nippoo:
I'm not sure if we want to include all those lines in `info` either. As a rule of thumb it sort of makes sense if `info` output is human-readable at the time - i.e. SpikeDetekt outputting a percentage progress message every 30 seconds makes sense, one `info` per cluster merge, etc. There's no way a human could keep up with the huge swathe of KK output at a rate of several lines per second. I agree it's important to save somewhere - to a file, in case it dies unexpectedly so you can debug it later, but humans are, at most, foing to want to see (in 'info')

- a graph / counter of iterations updated every minute or so (or even 5/10 minutes)
- maybe some statistic about the score delta to see how it's stabilising
- maybe the details of the last iteration / last 3 iterations should just be displayed and overwritten

You should be able to tail the log file if you want to see if it's getting stuck somewhere, or whatever, but feeding a ton of text to the browser doesn't seem ideal?

Perhaps we need to add a step in the logging hierarchy. I think some programs have "debug, notice, info, warn, critical" so we could have a "notice" which isn't logged, but save "info" for, say, a summary of every 100 iterations / 10 minutes (whichever comes first?).

rossant:
I agree that we should normalize conventions between phy and KK2. In phy, info is the standard level, it is used to show relevant information to the user in most cases. debug is used, well, for debug. :)

As a user, when running kk2 i want to see the current iteration, elapsed time, remaining time, maybe number of clusters, etc. We can also make a GUI wrapping this. All other information should be in debug mode confined to a file IMO.

thesamovar:
Normally (on large data sets) info level messages will pop out at a rate of about one every 10 minutes or so. It does spew out a lot of lines when attempting to split (I'll make an issue for that), but I think the rest of the lines (one per iteration) are important. @nsteinme what do you think about that?

However, if it's integrated into a GUI or an IPython notebook it would be fine to just show the last 5 or 10 lines of log output I guess? Again @nsteinme what do you think?

nsteinme:
Just read through this quickly, but yes I think seeing the last few info-level messages is useful, though possibly redundant with updating plots. Probably not necessary in any situation to show all info lines that have ever happened in the process - can always go to the klg for that.

thesamovar:
OK so maybe the best option is that in the gui we suppress all (or all but the last few messages if there is an easy way to do that) from the console log (although it will still go to a file), and have graphs that update after each iteration (typically every few minutes).

nsteinme:
That sounds good to me.

thesamovar:
OK I just pushed a change that suppresses some of the log messages. Info level logging should only be emitted a few times per iteration, or about once every few minutes on average in realistic data.

rossant:
@thesamovar @nippoo travis still unhappy because of cython dependency at runtime (should only be needed at build time)

thesamovar:
It's not using the latest version, neither the master version from github nor the latest release on pypi. All Cython and pyximport references are removed in the latest versions.

rossant:
it seems that `pip install klustakwik2` still installs the old version

thesamovar:
Any idea why? Is it because the release has dev in the name?

rossant:
you can log in on pypi and check that the new release is not hidden somehow

thesamovar:
It's not - I just downloaded it to check.

rossant:
does `pip install klustakwik2` install the right version for you?

thesamovar:
No, it gets 0.1.0 for me too...

thesamovar:
I think it is because of the dev in the name. If you ``pip install --pre klustakwik2`` it should work.

rossant:
travis still failing

```
Collecting klustakwik2
  Downloading klustakwik2-0.2.0.dev0.tar.bz2 (469kB)
    100% |████████████████████████████████| 471kB 511kB/s 
    Complete output from command python setup.py egg_info:
    Traceback (most recent call last):
      File "<string>", line 20, in <module>
      File "/tmp/pip-build-uY0vGD/klustakwik2/setup.py", line 20, in <module>
        from Cython.Build import cythonize
    ImportError: No module named Cython.Build
    
    ----------------------------------------
Command "python setup.py egg_info" failed with error code 1 in /tmp/pip-build-uY0vGD/klustakwik2
The command "pip install --pre klustakwik2" failed and exited with 1 during .
Your build has been stopped.
```

any ideas?

thesamovar:
This is a build-time dependency (it's in setup.py).

rossant:
travis is finally happy! thanks @thesamovar and @nippoo 
this is ready for review/merge

I've been doing a significant refactoring in the GUI code but I'll put it in another PR

thesamovar:
Great! I'll leave it to @nippoo to review/merge this.

PR #329, opened 2015-06-09
Refactor GUI

### LINKED ISSUE #72: Common interface for sorting algorithms
Inspired by scikit-learn:

### Spike detection
```
# We launch the spike detection.
# This will automatically use multiple CPUs if
# multiple engines have been launched with IPython.parallel.
# This call is asynchronous: the user can continue to work in the notebook,
# and request the task's status.
phy.spikedetect.run(model, algorithm="spikedetekt", ipp_view=c.load_balanced_view())

# Launch clustering.
phy.cluster.run(model, algorithm="klustakwik2", ipp_view=c.load_balanced_view())
```

### LINKED ISSUE #151: Correlograms normalization
Allow to switch between uniform normalization and per-row normalization.

### LINKED ISSUE #156: Ticks and baseline in CCG view
from @nsteinme :

> Would also be good to show a vertical line at some settable value, e.g. 3 ms, so it's easy to see if there are refractory violations. In the future it would be great to also show the results of some calculations related to this on these plots: for ACG, the estimated false positive rate (see Hill et al); for CCG, perhaps what the false positive rate would be if the two units were merged.

### LINKED ISSUE #211: Public functions for easily creating views
For example, `plot_waveforms(waveforms, channel_positions=..., ...)`

### LINKED ISSUE #250: Add no-cluster-store option
When opening an unclustered file with a single large cluster straight out of SpikeDetekt, or when opening a big file off a read-only filesystem, etc, we need to be able to open it up without having to create  a cluster store. This should only be the default when there exists only a single cluster in the chosen shank/clustering, but could be selected as an option if desired.

### LINKED ISSUE #272: Improve view testing
```python
@with_view
def test_my_view():
    c = MyView()
    yield  # show the view for n frames, n=2 by default
    c.x = y
    yield
```

* `n` should be settable from the command line when using `py.test`
* option from CLI to keep window interactive at the end of the function
* should work similarly with Qt widgets instead of VisPy views

### LINKED ISSUE #273: Improved similarity measure
https://github.com/kwikteam/phy/blob/master/phy/cluster/manual/session.py#L654

### LINKED ISSUE #302: Refactor tests with test case classes
To keep state between unit tests.

### LINKED ISSUE #305: More CLI options for `phy`
* `phy store [dir]`: show store status
  * more options: `clean`, `erase`
* `phy describe myfile.kwik`: describe data
* `phy notebook myfile.kwik`: create a new analysis notebook with the data open and open it
* refactor and use `argparse`
* tab completion
* ...

### LINKED ISSUE #306: Improve double-click zooming in grid view
The current implementation is a bit hackish and buggy. A few options

* Implement it more generally in `_panzoom`
* Implement it partially on the GPU
* Alternative: use a separate window with only 1 subplot, and bind it to double click (simpler implementation)

The tricky part in the implementation is to keep the state of all internal variables (including local pan zoom, constraints, etc) and reset them when the enlarge option is disabled.

@nsteinme what do you think of the alternative? basically you'd have two feature views, one 4x4 and one 1x1. By double clicking on a subplot it would zoom in the *other* view. This would make my life easier :) however it would take more place on your screen.

### LINKED ISSUE #320: Refactor session and GUI
```
Session
    model
    store (optional)
    settings
    
    open/close data
    change clustering/channel group
    
    view creator
        instanciate independent view models with the data
        use settings for views
        keep track of all created views
    gui creator
        instanciate guis with the data
        use settings for the gui
        keep track of all created guis
        view creator
    events
        open
        close


view creator (base)
    constructor
    name: view_model_class mapping
    create a view model with params settings
    save params when the view is closed


GUI (base)
    constructor
    view creator (by default, create one view of each type)
    dock window
    actions and shortcuts
    events
        close


GUI creator
    name: gui_class mapping
    create a GUI with parameters    
    save params when the GUI is closed
``` 

new module organization:

```
phy/cluster
    default_settings
    manual
        clustering
        gui (with ManualClusterGUI)
        static
            wizard.html
        wizard
phy/gui
    base (base GUI)
    dock
    qt
    session
    static  
        styles.css
        wrap_qt.html
    views (base view creator and gui creator)
```

### LINKED ISSUE #324: Show statistics of multiple clusters in stats view
Using an HTML table (row=stat, column=cluster)

### LINKED ISSUE #331: view.set_data
Useful to update a plot after it has been created

### LINKED ISSUE #333: Add clustering_name option in run_kk2

### LINKED ISSUE #334: set time as one dimension in the SingleFeatureView
It used to be "ctrl-t" in KV. 

For reference, I thought I would solve this problem myself by finding the object corresponding to that feature view, and calling some method to set it to have the time feature (which I suppose is probably feature #361). Here is a record of what I tried. 

I spent a while finding gui.views[3], which is the SingleFeatureViewModel, but I can't figure out how to call its "set_dimension" or "set_dimensions" methods without a crash, if those are even the right ones to call because I can't figure out how to see documentation for them (adding ? doesn't seem to work for this, i.e. gui.views[3].set_dimensions?) and I can't find where they are defined in the code (e.g. I looked into phy/plot/features.py, but this must not be the right thing because it suggests there should be a .features method which there isn't, and what the heck is "baking"??)). I also looked into the .view and .model parts of gui.views[3] and there's a lot of stuff in there and I have no idea what's going on. How does that relate to session.model, and what is a model, anyway? 

I'm sure this will become clearer when the documentation is up to date (many of the things there don't work anymore), so I can be patient until that happens, I understand it's not top priority which is fine. But if I should be thinking about things in a different way or if I am missing some resource that will help me understand what's going on, please let me know!

### LINKED ISSUE #337: Keybord shortcuts are hard to find
when pressing ctrl+h you only get Keyboard shortcuts for ClusterManualGUI, however to get the other shortcuts you have to mouse click on a specific window and press h

Opening test_hybrid_120sec.kwik
Running ubuntu 14.04

### LINKED ISSUE #339: Cannot reopen a dataset from Python 3 phy -> Python 2 phy
More generally we should test for this case... this was using a clusterstore created with phy/py3, now trying to reopen on phy/py2...

```
  File "c:\users\max\development\phydev\phy\phy\io\kwik\store_items.py", line 368, in __init__
    if self.disk_store else None)
  File "c:\users\max\development\phydev\phy\phy\io\store.py", line 227, in load_file
    return cPickle.load(f)
ValueError: unsupported pickle protocol: 3
```

### LINKED ISSUE #340: Global keyboard shortcuts (not focus-dependent)
We should really not have to focus a view to use its keyboard shortcuts - the entire clustering process should be doable with only a keyboard and no mouse.

We should have enough keys so as to be able to have global shortcuts, but otherwise we should be able to change focus from view to view using a single set of keys (maybe QWERTY or ASDFG or something).

On a related note, this is a good page: https://msdn.microsoft.com/en-us/library/ms971323.aspx#atg_keyboardshortcuts_keyboard_ui_design_rules

rossant:
* New generic `phy.gui` subpackage.
* All Qt utilities are now in `phy.gui`.
* There is a generic BaseSession that represents an interactive session around a Model (the data). The BaseSession handles:
  * open/close/save
  * user settings
  * GUI creation
* The BaseGUI allows to create a GUI around a main dock window with arbitrary widgets and keyboard shortcuts. The widgets are specified in the GUI as a `{name: view_model_class}` mapping. Other features include:
  * Add/close views
  * Save the view/GUI parameters in the internal settings file
  * Load keyboard shortcuts from user settings
  * Handle inter-view connections
* `phy.cluster.session` is now a lightweight session combining automatic, manual clustering, and potentially other workflows and GUIs. New GUIs can be defined interactively. In addition to the base settings' features, other features include:
  * change clustering interactively
  * change channel group
  * register custom statistic
* More fine-grained hierarchy of view model classes
* Better framework for Qt tests: one can now use a simple decorator to a test function and `yield` between programmatic GUI interactions. The delay between the interactions can be set from the command line with the `PHY_EVENT_LOOP_DELAY` env variable.
* Simplified cluster selection after wizard actions
* There are now two slightly different types of feature views: a grid view, and a single feature view showing just a single 2D plot. The dimensions in that plot can be changed from the grid view with ctrl+click.
* Tweaked keyboard shortcuts in the views.
* Added standalone plotting functions leveraging VisPy views (waveforms, CCGs, features, traces)
* Added customizable vertical lines in CCG view (allows to show refractory violations for example).
* Table showing the stats of all selected clusters in stats panel.
* Better default similarity measure.
* Added `session.cluster()` method to launch KK2.

## Closes

* closes #72
* closes #151 
* closes #156 
* closes #211 
* closes #250 
* closes #272 
* closes #273
* closes #302 
* closes #305 
* closes #306 
* closes #320 
* closes #324 
* closes #331 
* closes #333 
* closes #334 
* closes #337 
* closes #339 
* closes #340


rossant:

rossant:
@thesamovar I think i'll need a new KK2 0.2.1 release for travis to pass here

thesamovar:
Done! Probably need to restart your build.

rossant:
ok I think i've done everything I wanted to do in the PR
@nippoo can you review? the longer you wait the longer the PR risks to be :)

rossant:
I'll need to merge this today to go on with the integration of SpikeDetekt. @nippoo if you have further suggestions you can make them in a new issue/PR

phy/cluster/tests/test_view_models.py
```diff
@@ -82,7 +82,7 @@ def _test_empty(view_model_class, stop=True, **kwargs):
         return vm
 
 
-def _test_view_model(view_model_class, stop=True, use_store=True, **kwargs):
+def _test_view_model(view_model_class, stop=True, **kwargs):
```
nippoo: why no more `use_store`?

rossant: I think now a "mock" cluster store can be created if needed

PR #345, opened 2015-06-16
[MRG] Release tools

rossant:
Some automatic release tools. It's not completely done nor ideal but it's ready to merge: we can improve later, and I have some work to do before any release anyway...

Another PR will be uploaded once this one is merged.

### DIFF
--- a/tools/release.py
+++ b/tools/release.py
@@ -171,11 +171,4 @@ def release():
 
 
 if __name__ == '__main__':
-    if sys.argv[1] == 'release_test':
-        release_test()
-    elif sys.argv[1] == 'release':
-        release()
-    elif sys.argv[1] == 'docker_test':
-        _test_docker()
-    elif sys.argv[1] == 'docker_build':
-        _build_docker()
+    globals()[sys.argv[1]]()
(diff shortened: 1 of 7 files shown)


rossant:
ready to merge @nippoo

tools/release.py
```diff
@@ -0,0 +1,174 @@
...
+                                 )
+
+    release.upload_asset('application/zip', op.basename(path), path)
+
+
+def _git_commit(message, push=False):
+    assert message
+    if input("About to git commit {}: are you sure?") != 'yes':
+        return
+    _call('git commit -am "{}"'.format(message))
```
nippoo: maybe worth cd'ing to the current directory here? if you accidentally call it from another directory, it might commit and push to another repo!

rossant: yes, or we could also check the current git directory and raise an error if the name is not `phy` ..?

nippoo: Eh, yeah, that could work too! No need to over-think it (and some people might rename their directory or something). I'll merge for now.

rossant: is there any way to get the name of the git repo (if such thing exists? or the URL etc.), instead of the folder name?

nippoo: Yeah, you can
```
basename `git rev-parse --show-toplevel`
```
and check that == the package name...

PR #366, opened 2015-06-18
[MRG] Spike detection

### LINKED ISSUE #212: Spike sorting workflow in the notebook
Here is an example of how a complete spike sorting workflow, from a raw `.dat` file to a fully-sorted `.kwik file`, could look like from a user perspective:

```python
# We can provide a helper CLI tool to create a standard notebook.
# >>> phy myfile.dat  # creates the notebook and open it
# >>> phy myexp.kwik --spike-detection # creates a standard notebook for spike detection
# >>> phy myexp.kwik --automatic-clustering
# >>> phy myexp.kwik --manual-clustering

# Here is an example of a standard notebook starting from a .dat file.
import phy

# This memmaps the file (fast, nothing is loaded in RAM yet). Returns an object with memmapped 2D array, sample rate, and probe file as a nested dict.
# TODO: decide on the ideal probe file format and whether the channel mapping should be within the .prb or defined separately (i.e. does it often change for a given probe/lab)
data = phy.io.read_traces('myfile.dat', n_channels=120, sample_rate=20000, [probe='buszaki64.prb'], format=[flat, ns5, raw.kwd])

# This displays an interactive trace view.
phy.plot.plot_traces(data)

# If the user is happy with the data, create / save a KWIK file to contain all settings.
model = phy.io.create_kwik(prm)

from IPython.parallel import Client
c = Client()

# We launch the spike detection.
# This will automatically use multiple CPUs if
# multiple engines have been launched with IPython.parallel.
# This call is asynchronous: the user can continue to work in the notebook,
# and request the task's status.
phy.spikedetect.run(model, algorithm="spikedetekt", ipp_view=c.load_balanced_view())

# Launch clustering.
phy.cluster.run(model, algorithm="klustakwik2", ipp_view=c.load_balanced_view())

# Now we can start the manual clustering.
session = phy.cluster.manual.Session(model)
# ...
```

### LINKED ISSUE #217: Traces reader for .dat, .bin, .ns5, .raw.kwd. [...]
```
data = phy.io.read_traces('myfile.dat', n_channels=120, sample_rate=20000, [probe='buszaki64.prb'], format=[flat, ns5, raw.kwd])
```

Returns object with 2D memmapped array (`np.memmap`), probe file, and sample rate.

### LINKED ISSUE #220: Port SpikeDetekt2

### LINKED ISSUE #343: Whitening
from Pierre:

```python
def get_whitening_matrix(X, fudge=1E-18):
   from numpy import dot, sqrt, diag
   from numpy.linalg import eigh
   Xcov = dot(X.T, X)/X.shape[0]
   d,V  = eigh(Xcov)
   D    = diag(1./sqrt(d+fudge))
   W    = dot(dot(V,D),V.T)
   return W

# then, to whiten some C chunk:
C = dot(C, whitening) 
```

### LINKED ISSUE #347: lines should persist for all ccgs
I added lines with gui.views[4].view.lines = [-3, 0, 3] (great!) but when two clusters are selected, only the first panel shows the lines. They should all show them.

### LINKED ISSUE #348: problem switching features with waveform view?
In KV, it was the case that ctrl+left click on a channel in the waveform view made that channel's PC1 become the x-axis feature in feature view, whereas ctrl+right made it become the y-axis feature. Now, I think ctrl+left is doing the same thing, but I'm not sure what ctrl+right is doing. The problem is that if you use the "t" feature to make the x-dimension time, there is no way to see PC1 of a desired channel on the y-dimension. 

Relatedly, is there a way to choose PC2 or PC3? 

And, it would be helpful at some point to have the kind of axis labeling that KV had (nothing more complicated than a channel number and feature number on each axis, would be fine.

### LINKED ISSUE #351: on_view_open() in user_settings
Allows to modify a view when it is opened.

### LINKED ISSUE #352: Log file

### LINKED ISSUE #356: Prompt save dialog when exiting the application with the desktop manager
Currently, the prompt is only displayed when using `ctrl+q`.

### LINKED ISSUE #360: Improve default dimensions in feature grid view
@nsteinme what should be the `x` and `y` dimensions in subplot `(i, j)`? Previously, `x=dim_j, y=dim_i`, now this is reversed. I can find arguments for both ways. WDYT?

Also, maybe we could show different dimensions in `(i, j)` and `(j, i)` (currently, the plots are just symmetric)

### LINKED ISSUE #364: Add cluster_ids option in cluster-manual CLI tool
To open the GUI on a set of clusters

### LINKED ISSUE #365: Custom dimension in feature view
something along the lines of:

```python
feature_view.add_extra_feature('my_new_feature', n_spikes_long_array)
feature_view.set_dimensions('time', 'my_new_feature')
```

### LINKED ISSUE #369: Don't create the store in session if there is only one cluster

### LINKED ISSUE #370: Bring back to life 1+click shortcut

### LINKED ISSUE #371: Scale factor in feature view

rossant:
This PR brings SpikeDetekt to phy (1 CPU only for now).
- Individual spike detection processing algorithms in `phy.traces`
- Spike detection main class in `phy.cluster.algorithms`
- Improvements to the feature view
- Improved HTML view models
- Added a file logger
- Clearer color map
- Refactored feature view dimensions
- `KwikCreator` creates Kwik files.
- Load raw data from `.kwd` or `.dat` files in `KwikModel`
- Multi-shank capabilities in `MEA` class
- `session.detect_spikes()` can be used to launch spike detection on the data
- Bug fixes
### Closes
- Closes #212
- Closes #217
- Closes #220
- Closes #343
- Closes #347
- Closes #348
- Closes #351
- Closes #352
- Closes #356
- Closes #360
- Closes #364
- Closes #365
- Closes #369
- Closes #370
- Closes #371


rossant:
rebased

rossant:
@nippoo you can start reviewing this

rossant:
rebased done

phy/cluster/algorithms.py
```diff
@@ -0,0 +1,671 @@
...
+        with open(path, 'rb') as f:
+            return _load_ndarray(f, dtype=dtype, shape=shape)
+
+    def _load_data_chunks(self, name,
+                          n_samples=None,
+                          n_channels=None,
+                          groups=None,
+                          spike_counts=None,
+                          ):
+        _, _, keys, _ = zip(*list(self.iter_chunks(n_samples, n_channels)))
```
nippoo: this is one of my least favourite bits of Python...

PR #397, opened 2015-06-24
[MRG] Bug fixes

### LINKED ISSUE #216: Create empty KWIK from PRM file
PRM file loader:

```
model = phy.io.create_kwik(prm)
```

### LINKED ISSUE #355: split with lasso doesn't seem to work properly
Here is my lasso:

image

And here is what I see after I press k:

image

### LINKED ISSUE #358: higher contrast colors?
Very minor point, but the default green is low contrast against the grey, e.g. I have a lot of trouble seeing which points are green vs. grey in this display, especially on the left where they are sparse: 

image

And given that there is no option to change cluster colors, it's something we should get right. Something with max saturation should do?

### LINKED ISSUE #375: Simplify dev tools in setup.py
see what was done at https://github.com/rossant/ipymd/pull/55/files

### LINKED ISSUE #393: Create empty user_settings file

### LINKED ISSUE #395: Update dictionaries in settings
currently, defining a dictionary as a user setting overrides the default dictionary, while it should be updated

rossant:
* Various bug fixes
* Better handling of dictionary settings
* Snippet commands

## Closes

* Closes #216
* Closes #355
* Closes #358
* Closes #375
* Closes #393
* Closes #395

### DIFF
--- a/phy/cluster/algorithms.py
+++ b/phy/cluster/algorithms.py
@@ -36,7 +36,7 @@ def _keep_spikes(samples, bounds):
 
 def _split_spikes(groups, idx=None, **arrs):
     """Split spike data according to the channel group."""
-    dtypes = {'spike_samples': np.uint64,
+    dtypes = {'spike_samples': np.float64,
               'waveforms': np.float32,
               'masks': np.float32,
               }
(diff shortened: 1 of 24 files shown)


rossant:
@nippoo ready for review

phy/cluster/algorithms.py
```diff
@@ -36,7 +36,7 @@ def _keep_spikes(samples, bounds):
 
 def _split_spikes(groups, idx=None, **arrs):
     """Split spike data according to the channel group."""
-    dtypes = {'spike_samples': np.uint64,
+    dtypes = {'spike_samples': np.float64,
```
rossant: this is to keep the fractional part and save it later to the Kwik fiel

PR #417, opened 2015-06-25
[MRG] Show git version with fallback. Closes #410

### LINKED ISSUE #410: Show the full tag/git version in `phy -v`

nippoo:


phy/__init__.py
```diff
@@ -26,7 +27,8 @@
 
 __author__ = 'Kwik team'
 __email__ = 'cyrille.rossant at gmail.com'
-__version__ = '0.2.0.dev0'
+__version__ = '0.2.0-dev1'
```
rossant: should be dev0 as long as we have not done a pypi release...

phy/utils/_misc.py
```diff
@@ -98,3 +100,17 @@ def _show_shortcuts(shortcuts, name=''):
     for name in sorted(shortcuts):
         print('{0:<40}: {1:s}'.format(name, _show_shortcut(shortcuts[name])))
     print()
+
+def _git_version():
```
rossant: can you add a unit test for this function?

nippoo: I'm not sure really. What would you suggest? This sort of relies on having command-line git installed, and it will basically return a random string... the test would basically look like the function itself?...

phy/__init__.py
nippoo: are we doing pypi dev releases?

phy/utils/_misc.py
rossant: well git is installed on travis (or we can easily install it), so you can just call this function and check that it returns a string with the expected number of characters...

phy/utils/_misc.py
```diff
@@ -98,3 +100,17 @@ def _show_shortcuts(shortcuts, name=''):
...
+    curdir = os.getcwd()
+    filedir, _ = op.split(__file__)
+    os.chdir(filedir)
+    try:
+        fnull = open(os.devnull, 'w')
+        version = 'git-'+subprocess.check_output(['git', 'describe', '--abbrev=8', '--dirty',
+                                                       '--always', '--tags'], stderr=fnull).strip().decode('ascii')
+        return version
+    except:
```
rossant: should except one or several expected exceptions (that's for your diction)

phy/utils/_misc.py
```diff
@@ -98,3 +100,17 @@ def _show_shortcuts(shortcuts, name=''):
...
         print('{0:<40}: {1:s}'.format(name, _show_shortcut(shortcuts[name])))
     print()
+
+def _git_version():
+    curdir = os.getcwd()
+    filedir, _ = op.split(__file__)
+    os.chdir(filedir)
+    try:
+        fnull = open(os.devnull, 'w')
+        version = 'git-'+subprocess.check_output(['git', 'describe', '--abbrev=8', '--dirty',
```
rossant: spaces between the +

phy/utils/_misc.py
```diff
@@ -98,3 +100,17 @@ def _show_shortcuts(shortcuts, name=''):
     for name in sorted(shortcuts):
         print('{0:<40}: {1:s}'.format(name, _show_shortcut(shortcuts[name])))
     print()
+
+def _git_version():
+    curdir = os.getcwd()
+    filedir, _ = op.split(__file__)
+    os.chdir(filedir)
```
rossant: maybe use http://stackoverflow.com/a/20115526/1595060 instead of chdir

phy/utils/_misc.py
```diff
@@ -98,3 +100,17 @@ def _show_shortcuts(shortcuts, name=''):
     for name in sorted(shortcuts):
         print('{0:<40}: {1:s}'.format(name, _show_shortcut(shortcuts[name])))
     print()
+
```
rossant: need an extra space

phy/__init__.py
rossant: yes

phy/utils/_misc.py
nippoo: ok! but this does mean that all the tests will fail if someone doesn't have git installed, or have downloaded the package rather than cloning the repository (even if they do have git installed).

rossant: maybe add a `warn()` if the function returns None instead of failing..?

rossant:
looks like this needs a rebase @nippoo

phy/__init__.py
```diff
@@ -26,7 +27,8 @@
 
 __author__ = 'Kwik team'
 __email__ = 'cyrille.rossant at gmail.com'
-__version__ = '0.2.0.dev0'
+__version__ = '0.2.0-dev0'
```
rossant: I think this needs to be reverted: https://www.python.org/dev/peps/pep-0440/#public-version-identifiers

rossant:
python 2 fail: `ameError: global name 'WindowsError' is not defined`

phy/__init__.py
```diff
@@ -27,6 +28,7 @@
 __author__ = 'Kwik team'
 __email__ = 'cyrille.rossant at gmail.com'
 __version__ = '0.2.0.dev0'
+__version_git__ = _git_version() or __version__
```
rossant: so I guess we'll just have to change it here?

nippoo:
@rossant this is good to go!

PR #418, opened 2015-06-25
Improve phy command-line tool and spike detection

### LINKED ISSUE #214: Integrate a set of common probes
* Tetrode
* 8-channel
* 32-channel
* ...

### LINKED ISSUE #378: Use subparsers in CLI tool
cleaner code to handle subcommands with `phy` CLI tool
https://docs.python.org/dev/library/argparse.html#argparse.ArgumentParser.add_subparsers

### LINKED ISSUE #394: Keyboard shortcut "snippets" in GUI
So this would be a flexible, easy-to-use, powerful interface to change parameters without having dedicated keyboard shortcuts. Inspired by vim's keyboard shortcuts... the goal would not be to replace IPython for complex or one-off tasks, but provide a fast and powerful interface of performing common tasks within the GUI without touching the mouse or switching focus.

Examples:
- change correlogram width (`:cv w 100` displays 100ms width) or bin size (`:cv bin 10` = 10ms bin size) or both (`:cv bin 10 w 100`) or number of spikes `:cv spikes 1000`
- skip to traceview time (`:tv 1000` skips to 1000 seconds or `:tv 1000 1001` shows that time duration or something)
- select clusters (`:c 1004` or `:c 1004 23 430 2101` or `:c 1000 to 1023` or `:c Nick Alex Cyrille Bob Alice Kenneth` or something)
- select featureview features (`:fv 9.1 10.3 10.1 t` selects channel 9 feature 1, channel 10 feature 3, channel 10 feature 1, time)
- maybe waveforms/features visible `:fv bg 100000` or `:fv fg 10000` (featureview foreground/background 100000) or `:fv bg default` or `:wv fg 1000` (waveformview foreground 1000 spikes) 

to quote @rossant:
* something like `:c 10 20<enter>` shows those 2 clusters. when you press `:`, you're in "snippet mode", and everything you type goes to the status bar. if you type backspace, it erases a character as expected. press enter to execute the snippet and clear the status bar. also, press `esc` to leave the snippet mode.
* the functionality would be built into BaseGUI, which ClusterManualGUI derives from.
* the list of snippets is completely extensible, and the default ones could be part of `default_settings.py`. a snippet is defined by a key-value pair where the key is a regex, and the value is a function that takes as argument the gui and the regex output
* example

```python

def select_clusters(gui, r):
    cluster = r.group(1)
    gui.select(cluster)

cluster_manual_snippets = {
  r'c ([\d]+)': select_clusters,
  # ...
}
```

* you can also add snippets interactively:

```python
@gui.register_snippet(r'c ([\d]+)')
def select_clusters(gui, r):
    pass
```

thoughts @nsteinme ?

### LINKED ISSUE #403: phy detect params.prm
This is mostly to simplify the migration from klusta to phy in the short term. The command just runs spike detection.

### LINKED ISSUE #404: Test spike detection on a real dataset

### LINKED ISSUE #412: snippets not working on windows under pr/397
Console reports entering snippet mode when I type ":", but no commands seem to work, and now things like spacebar to change cluster don't work, even after pressing escape many times.

### LINKED ISSUE #413: after merging two clusters, pressing <enter> should pin the newly created (and now selected) cluster
After merging the new cluster is correctly displayed and seems to be selected. If you press enter here you'd expect to pin it and continue working on that cluster. 

Currently, instead, it pins the next one in "spacebar order", i.e. the next one you'd see if you pressed spacebar rather than enter.

### LINKED ISSUE #416: Limit extra_features and cluster statistics to a certain number of sf...
aka 2/3 shouldn't be 0.666666666666666666666667 ;)

### LINKED ISSUE #419: Progress report during spike detection in CLI tool

### LINKED ISSUE #421: cutoff in time view
it seems only a subrange of time values are plotted in the feature views for some clusters. for example the cluster below, before and after splitting:
time_cutoff
before splitting, there are no points after a certain time, after splitting you can see that there were in fact spikes across the whole recording.

### LINKED ISSUE #430: Show cluster group in stats panel

### LINKED ISSUE #432: fails poorly when raw.kwd not present
I thought this was fixed before(?), but I get a million "error while loading waveform: invalid time" errors if the raw.kwd is not present, rather than being reminded about what I did wrong.

### LINKED ISSUE #435: Do not save files in download_data if the URLs don't exist

### LINKED ISSUE #439: Improve wizard selection after split
The new best cluster should be the one just before the cluster that was split.

### LINKED ISSUE #441: error when ctrl+right click in waveform view after setting feature dimension "manually"
When I set the y-dimension in feature view to be depth (via view.set_dimension), then I try ctrl+right click on waveform view (which should change the y-dim to be pc1 of that channel), it gives this error:

image

rossant:
This PR improves the `phy` script and fixes several bugs.

* Use subparsers in `phy` CLI tool
* `phy traces` CLI script to visualize raw traces
* Show cluster group in stats panel
* Add library of probe files
* Bug fixes

## Closes

* Closes #214 
* Closes #378 
* Closes #394 
* Closes #403 
* Closes #404 
* Closes #412 
* Closes #413 
* Closes #416
* Closes #419
* Closes #421
* Closes #430
* Closes #432
* Closes #435
* Closes #439 
* Closes #441


rossant:

phy/cluster/algorithms.py
```diff
@@ -67,37 +68,76 @@ def _array_list(arrs):
 
 
 def _concat(arr, dtype=None):
-    return np.array([_[...] for _ in arr], dtype=dtype)
+    out = np.array([_[...] for _ in arr], dtype=dtype)
+    return out
```
nippoo: why?

rossant: hahaha.. it's a debugging leftover..

phy/cluster/algorithms.py
```diff
@@ -405,13 +464,18 @@ def step_extract(self, bounds, components,
...
+                             shape=(n_samples, n_channels), key=key,
+                             lazy=False,
+                             )
         # Extract the spikes from the chunk.
         groups, spike_samples, waveforms, masks = self.extract_spikes(
             components, chunk_f, thresholds=thresholds)
 
         # Remove spikes in the overlapping bands.
-        idx = _keep_spikes(spike_samples, (keep_start, keep_end))
+        # WANRING: add keep_start to spike_samples, because spike_samples
```
nippoo: I LOVE WANRINGS

phy/scripts/phy_script.py
```diff
@@ -32,169 +39,250 @@ def error(self, message):
...
+    from vispy.app import run
+    from phy.plot.traces import TraceView
+    from phy.io.h5 import open_h5
+    from phy.io.traces import read_kwd, read_dat
+
+    path = args.file
+    if path.endswith('.kwd'):
+        f = open_h5(args.file)
+        traces = read_kwd(f)
+    elif path.endswith(('.dat', '.bin')):
```
nippoo: there should be an else: "filetype not supported, please rename /..."

also we should make sure we're not trying to read/write to the same KWD file...!

PR #443, opened 2015-06-29
Fixing the mean_probe_position element in store

### LINKED ISSUE #440: mean_probe_position incorrect
Right now you are calculating the position of the cluster as:
mean ( masks * probe.positions )

But it should be:
sum ( masks * probe.positions ) / sum(masks)

[that's just the formula for center of mass]

Here's a specific line that computes it for each spike individually; you can probably use it directly if masks is the same size? I'm not sure.

masks = view.model.masks[...]
x = np.sum(masks*view.model.probe.positions[:,1],axis=1)/np.sum(masks,axis=1)

nsteinme:
This fix returns the center of mass in the "y" dimension.

Closes #440

### DIFF
--- a/phy/io/kwik/store_items.py
+++ b/phy/io/kwik/store_items.py
@@ -51,7 +51,6 @@ def _mean(arr, shape):
             return arr.mean(axis=0)
     return np.zeros(shape, dtype=np.float32)
 
-	
 
 #------------------------------------------------------------------------------
 # Store items
@@ -532,12 +531,14 @@ class ClusterStatistics(FixedSizeItem):
     def mean_probe_position(self, cluster):
         mean_masks = self.load(cluster, 'mean_masks')
         if mean_masks is not None and mean_masks.shape[0]:
-            meanClusterPosition = np.sum(self.model.probe.positions * mean_masks[:, np.newaxis], axis=0) / max(1, np.sum(mean_masks))
+            meanClusterPosition = (np.sum(self.model.probe.positions *
+                                   mean_masks[:, np.newaxis], axis=0)
+                                   / max(1, np.sum(mean_masks)))
         else:
             meanClusterPosition = np.zeros((2,), dtype=np.float32)
-            
+
         return meanClusterPosition
-        
+
     def n_spikes(self, cluster):
         return len(self._spikes_per_cluster[cluster])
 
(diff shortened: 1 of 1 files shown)


phy/io/kwik/store_items.py
```diff
@@ -528,8 +528,8 @@ def n_unmasked_channels(self, cluster):
 
     def mean_probe_position(self, cluster):
         mean_masks = self.load(cluster, 'mean_masks')
-        return _mean(self.model.probe.positions *
-                     mean_masks[:, np.newaxis], (2,))
+        return np.sum(self.model.probe.positions[:, 1] *
+                      mean_masks) / np.sum(mean_masks)
```
rossant: * this should return the mean position with two coordinates x and y, not just y
* you should create a new `_sum()` function, like the `_mean` one, to deal with empty arrays

phy/io/kwik/store_items.py
```diff
@@ -528,8 +528,14 @@ def n_unmasked_channels(self, cluster):
 
     def mean_probe_position(self, cluster):
         mean_masks = self.load(cluster, 'mean_masks')
-        return _mean(self.model.probe.positions *
-                     mean_masks[:, np.newaxis], (2,))
+        if mean_masks is not None and mean_masks.shape[0]:
+            meanClusterPosition = (np.sum(self.model.probe.positions *
```
rossant: could you rename this to `mean_cluster_position`? This is to stay consistent with the rest of the code and with standard Python conventions

rossant:
I suppose you need to rebase for tests to pass (I think they were broken on master...) ?

something like this should work (only if upstream=kwikteam and origin=nsteinme, cc @nippoo)

```
git checkout -b backup
git checkout fix-mean-probe
git fetch --all
git rebase upstream/master
git push origin --force
```

nsteinme:
Looks like the test that failed is that it wasn't able to download some
sample data? Anyway, ok, I can try your think, let's see...

nsteinme:
Ok, I did that. Do I need to make it into a new pr or something now? Or is it good to go?

rossant:
perfect! let's see what travis thinks...

rossant:
Travis is happy, and this seems to work on the data; merging.
Thanks @nsteinme for your efforts! :)

nsteinme:
Woohoo! And in the process I almost passed @nippoo for number of commits ;)

PR #448, opened 2015-06-29
Bug fixes in spike detection

### LINKED ISSUE #445: Handle KlustaKwik2 parameters
* How to specify them? In the `prm` file? (for 0.2.0 we have `phy spikesort` that acts as a drop-in replacement for `klusta`) 
* Should we maintain strict compatibility with the old PRM files? 
* Have KK2 parameter names and values changed compared to KK1?
* What should be the default parameter of `num_starting_clusters`?
* Should we rename `num_starting_clusters` to `n_starting_clusters` (at least in phy) for  consistency with all other parameters in phy?

cc @nippoo @thesamovar

### LINKED ISSUE #446: wizard not skipping MUA and Good clusters properly
When I assign a cluster to MUA or Good (didn't test this with Noise), it gets the correct label but doesn't move its position in the wizard list (or get removed from the list). This means that it gets pinned (as "best unsorted") when you classify the one previous to it in the list. Probably it should pick a new best unsorted as - the next one in the list that has cluster_group=Unsorted.

rossant:
* New `phy spikesort` tool, replacing `klusta`
* Improved progress reporter
* Take dead channels into account during spike detection
* No more memory mapped files in store by default, but might need to check RAM consumption (?)
* Update the `PRM` format to now have `spikedetekt` and `klustakwik2` dictionary parameters

## Closes

* Closes #445 
* Closes #446


rossant:
@nsteinme this is the PR I used to run `phy spikesort` on your data

rossant:
rebased

nsteinme:
Just to be clear, are you saying you'd like me to try it out now? If so,
what kind of command do I issue? Do I need to make a new prm file first?

rossant:
well, yes you can if you want, without any changes to the PRM for now, because I haven't implement the PRM changes yet.

phy/cluster/algorithms.py
```diff
@@ -155,7 +164,7 @@ def __init__(self, tempdir=None, probe=None, **kwargs):
     def _create_filter(self):
         rate = self._kwargs['sample_rate']
         low = self._kwargs['filter_low']
```
nippoo: Shouldn't this be `* rate` too?

rossant: I don't think so

nippoo: why is the low freq specified in samples and the high freq in seconds?... that seems odd

rossant: that's always been the case. I assume these parameters do not normally depend on the sample rate, but for the high frequency, you want to set the high one to the max, i.e. about half of the sample rate or something... @shabnamkadir ?

nippoo: Not saying this wasn't the case before, but it seems rather inconsistent...?

rossant: well I'd say these two parameters are less similar than what their names imply, and they play different roles... but @shabnamkadir will know more.

nsteinme: The previous default was this:

# Bandpass filter low corner frequency
filter_low = 500.

# Bandpass filter high corner frequency
filter_high = 0.95 * .5 * sample_rate

And the units on both of these are Hz (1/sec). Neither are in samples or
seconds! The second one is computed from the sample_rate (which also has
units of Hz) to be just below Nyquist, that's all. In fact I guess this
filter_high part is filtering almost nothing out - that's really the
low-pass filter you'd need on your hardware before the signal was
digitized. You might as well just be using a high-pass at this point rather
than bandpass.

While we're here, is there a reason that you don't get to specify the
filter that's used to make the LFP [low.kwd]? This should probably be a
parameter too.

Nick

nippoo: I agree - both should be in Hz.

For the .low.kwd - I don't think we need to extract LFP at the time of spike detection. We should be able to do this later, at the time of analysis.

nsteinme: Disagree - it takes a long time to even just load the LFP from disk (last
night, e.g., the one I was working with was 5000000x129 (int16) = 1.5GB).
You definitely don't want to load the 16x larger raw data and compute the
LFP from it when going to analyze that. If you're already loading the raw
data to detect spikes, this seems like a fine time to filter and save the
LFP.

rossant: @nippoo I had misread your comment; yes these parameters have always been in Hertz!

We can save the LFP, but in which format? We already save the HPF in temporary files (flat binary), one file per chunk (to avoid contention issues), shall we do the same for LFP too? In any case @nsteinme please add an issue otherwise we'll forget it...

rossant:

phy/cluster/algorithms.py
nsteinme: Sorry, I may have missed something, are you doing away with the .low.kwd
file? I'm not talking about saving the LFP in the store, but rather in a
file like low.kwd.

Will add issue on specifying LFP filtering parameters. Is this a phy issue
now or spikedetekt?

rossant: We want to get rid of all `.kwd` files. Issues are all in `phy` now.

nippoo: A phy issue - spikedetekt is no more.

I think we're doing away with the concept of .kwd files altogether. You provide your raw data file in .dat/bin/ns5/whatever, then everything is read in realtime from this file. High-pass filtered data aren't saved, they're simply processed and features extracted as needed. Waveforms are extracted, filtered and aligned in realtime every time you load a cluster. We could write a function to low-pass the data, and save it to a new .dat or NumPy array, and if it ends up being really slow (which it shouldn't?) then we could do this at SpikeDetekt-time, but my preference would be to keep the processes separate and modular, and let people extract/save LFP with customisable parameters at some later stage (normally during analysis after clustering)...

rossant: > High-pass filtered data aren't saved, they're simply processed and features extracted as needed. Waveforms are extracted, filtered and aligned in realtime every time you load a cluster.

Actually, these *are* all saved in the temporary store, notably for performance reasons. All extracted waveforms are saved on disk during spike detection, and, currently, they stay here until you delete them. In the GUI, the subset of waveforms of every cluster are also saved when you first load your dataset in phy.

Regarding LFP, I'd rather save them *during* spike detection actually, because we can benefit from the fact that the raw data is already loaded in RAM. This will also be nicely parallelized once SD is parallelized.

nsteinme: All right, sounds like it's empirical. If you can prove to me that getting
a channel's LFP is nearly as fast from the raw data as from being read from
a dedicated file, then fine :)

rossant: @nsteinme what analyses do you plan to do on LFP data? can these analyses be done chunk by chunk?

nsteinme: Sometimes, e.g. viewing a segment of the data during a particular time
window (e.g. trial), or computing an event-triggered average of the LFP for
an infrequent event. But then sometimes you'd want to compute a power
spectrum across many segments of data, or coherence between two LFP traces
across the whole data, or spike-triggered average (which would involve the
whole data trace assuming spikes are high enough rate).

thesamovar:
I changed KK2 so that now you don't pass an argument to ``cluster_mask_starts`` and the ``num_starting_clusters`` is a standard KK2 parameter like any of the others.

rossant:
thanks @thesamovar! can you make a new pip release so that travis can pass here?

rossant:
thanks @thesamovar! can you make a new pip release so that travis can pass here?

phy/cluster/algorithms.py
```diff
@@ -18,7 +18,9 @@
 from ..utils._types import Bunch
 from ..utils.event import EventEmitter, ProgressReporter
 from ..utils.logging import debug, info
-from ..electrode.mea import _channels_per_group, _probe_adjacency_list
+from ..electrode.mea import (_channels_per_group,
```
nippoo: is there any reason for this, out of interest? or just a leftover from adding and removing some import?

rossant: oops, leftover..

phy/cluster/algorithms.py
nippoo: Uhhhh.

Un-sidetracking this discussion a bit...

Nick, should both these be frequencies in Hz?

This is my main objection before merging!

rossant: I realize that I forgot to explain the motivation for the change: having a `sample_rate` variable in the parameter files makes the code much more complicated.  You can't have a simple dictionary for the parameters, you need a function of the `sample_rate` variable. That's why I switched to a factor of the sample rate.

thesamovar:
Done!

rossant:
thanks! travis happy :)

phy/cluster/algorithms.py
```diff
@@ -245,6 +257,14 @@ def detect(self, traces_f, thresholds=None):
         # Compute the threshold crossings.
         weak = thresholder.detect(traces_t, 'weak')
         strong = thresholder.detect(traces_t, 'strong')
+        # Force crossings to be False on dead channels.
```
nippoo: This means that the graph has to be manually regenerated, yes?

rossant: it can, yeah

phy/cluster/default_settings.py
```diff
@@ -7,47 +7,48 @@
...
-        filter_low=500.,
-        filter_high=0.95 * .5 * sample_rate,
-        filter_butter_order=3,
 
-        # Data chunks.
-        chunk_size=int(1. * sample_rate),
-        chunk_overlap=int(.015 * sample_rate),
+spikedetekt = {
+    'filter_low': 500.,
+    'filter_high_factor': 0.95 * .5,  # will be multiplied by the sample rate
```
nippoo: Aha, now I understand. This is simply an anti-aliasing filter.

Could we instead rename this parameter to aa_nyquist_filter_factor to make its purpose more explicit? Confusingly it's a lowpass filter. 

In fact I would personally vote to have this as a fixed frequency (0.47*sample_rate) and just set one variable aa_filter = True/False - why would we really want to change this? @nsteinme

phy/io/kwik/creator.py
```diff
@@ -297,9 +299,18 @@ def add_clustering(self,
...
 
 
-def create_kwik(prm_file=None, kwik_path=None, probe=None, **kwargs):
+def create_kwik(prm_file=None, kwik_path=None, overwrite=False,
+                probe=None, **kwargs):
     prm = _read_python(prm_file) if prm_file else {}
-    sample_rate = kwargs.get('sample_rate', prm.get('sample_rate'))
+
+    if prm:
+        assert 'spikedetekt' in prm
```
nippoo: todo: we should be able to convert to .kwik without running spike detection (i.e. separate those two processes, and manually specify / detect spikes using another algorithm)

rossant: I think that's already possible with kwik creator and add_spikes

phy/io/store.py
```diff
@@ -51,7 +51,9 @@ def _default_array(shape, value=0, n_spikes=0, dtype=np.float32):
 def _assert_per_cluster_data_compatible(d_0, d_1):
     n_0 = {k: len(v) for (k, v) in d_0.items()}
     n_1 = {k: len(v) for (k, v) in d_1.items()}
-    assert n_0 == n_1
+    if n_0 != n_1:
+        raise IOError("Inconsistency in the cluster store: please remove "
```
nippoo: we could do this automatically?

phy/scripts/phy_script.py
```diff
@@ -45,6 +45,8 @@ def error(self, message):
   phy -v                display the version of phy
   phy describe my_file.kwik
                         display information about a Kwik dataset
+  phy spikesort my_params.prm
+                        run the whole suite (spike detection and clustering)
   phy detect my_params.prm
                         run spike detection on a parameters file
   phy cluster-auto my_file.kwik --num-clusters-start=100
```
nippoo: Presumably this will be `phy cluster-auto myfile.kwik my_params.prm` not just `myfile.kwik`? How else are you going to specify the parameters?

Could also have the option for `phy cluster-audio myfile.kwik --algorithm=klustakwik2 --parameter1=blob --parameter2=blab` I guess

phy/utils/event.py
```diff
@@ -166,16 +166,18 @@ def __init__(self):
...
         The string needs to contain `{progress}`.
 
         """
 
+        end = '\r' if not line_break else None
+
         @self.connect
         def on_progress(value, value_max, **kwargs):
-            _default_on_progress(message, value, value_max, **kwargs)
+            _default_on_progress(message, value, value_max, end=end, **kwargs)
```
nippoo: We should check it contains `{progress}`

phy/cluster/default_settings.py
nsteinme: I don't know under what situation you would want to change that filter.
Might have been a parameter for a good reason before, though. Ask Kenneth?

kdharris101: No. In fact, I don’t think this is necessary at all. It has always been there, since Jozsef Csicsvari’s original code, but I have never seen the point in using a bandpass rather than simply a highpass filter.

rossant: > No. In fact, I don’t think this is necessary at all. It has always been there, since Jozsef Csicsvari’s original code, but I have never seen the point in using a bandpass rather than simply a highpass filter.

OK -- considering that for the next release...

nippoo: Oh - @kdharris101 - you'd suggest omitting the LPF altogether? My understanding is that it was to eliminate aliasing artefacts, but on second thoughts putting that after the ADC stage makes no sense...

phy/io/store.py
rossant: yeah, we should, maybe catching this in the session when creating the store

phy/scripts/phy_script.py
rossant: that option would be good. In any case I suppose most users will just use `phy spikesort` instead

PR #477, opened 2015-07-02
Reverted commit 9433b92 to support git pre-1.8

nippoo:


nippoo:
@rossant good to merge from my end

PR #493, opened 2015-07-02
[MRG] Fix `phy cluster-auto`  command

### LINKED ISSUE #485: phy cluster-auto should take a PRM file

rossant:
Closes #485.

### DIFF
--- a/phy/scripts/phy_script.py
+++ b/phy/scripts/phy_script.py
@@ -299,7 +299,7 @@ def cluster_auto(args):
     ns = dict(session=session,
               clustering=args.clustering,
               )
-    cmd = ('session.cluster(clustering=clustering, save_kk2_iter=False)')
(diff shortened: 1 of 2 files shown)


rossant:
status: waiting to fix KK2 split problem with callback

rossant:
ready to go @nippoo

phy/scripts/phy_script.py
```diff
@@ -294,11 +294,11 @@ def detect(args):
 
 
 def cluster_auto(args):
+    assert args.file.endswith('.kwik')
```
nippoo: `assert`? this should probably be in `_create_session`, no? And also fail with a more sensible error than just an `AssertError`?

I can imagine some people renaming their .prm to a .kwik ;-)

rossant: yeah... the `phy_script` module needs a refactoring too...

PR #495, opened 2015-07-03
[MRG] Win64 fixes

rossant:
Trying to make phy work with h5py 2.5.0 on win64 (workaround the segmentation faults)


rossant:
ready to go @nippoo @nsteinme 
you might want to double-check that you're able to run `phy spikesort` on Windows on a small chunk of data (`--interval`). It should work with h5py 2.5.0 now.

PR #502, opened 2015-07-04
[MRG] Use JSON for internal settings

### LINKED ISSUE #484: Switch to JSON for internal settings

rossant:
* Closes #484.
* Drop Python 3.3 support, because of a bug on travis with bytes, string, and encodings...


rossant:
ready to merge @nippoo

PR #505, opened 2015-07-05
[MRG] Add phy/gui/static files

rossant:
Fix bug with `phy/gui/static/*.html` files not installed.


rossant:
Ready to merge @nippoo

PR #512, opened 2015-07-06
[MRG] Fix multiple thresholds

### LINKED ISSUE #496: Fix bug with single_threshold=False

rossant:
Closes #496

### DIFF
--- a/phy/traces/detect.py
+++ b/phy/traces/detect.py
@@ -70,7 +70,7 @@ class Thresholder(object):
                  thresholds=None,
                  ):
         assert mode in ('positive', 'negative', 'both')
-        if isinstance(thresholds, (float, int)):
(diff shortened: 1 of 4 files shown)


rossant:
ready to merge @nippoo

phy/traces/detect.py
```diff
@@ -99,7 +99,7 @@ def detect(self, data_t, threshold=None):
         if isinstance(threshold, string_types):
             assert threshold in self._thresholds
             threshold = self._thresholds[threshold]
-        threshold = float(threshold)
+        # threshold = float(threshold)
```
nippoo: why is this now commented?

rossant: * bug when this is an array
* not sure why I needed this!
* should actually be removed!

PR #513, opened 2015-07-06
[MRG] Skip KK2 split iterations in event emitter

rossant:


rossant:
good to merge @nippoo

Issue #462, opened 2015-07-01
Improve dev tools


nippoo:
- [x] `pip install -r requirements-dev.txt` vs `pip install -r requirements-travis.txt`
- [x] remove dependency on `make` by adding custom distutils commands http://stackoverflow.com/questions/1710839/custom-distutils-commands/1712544#1712544

rossant:
keep 0.2.0 milestone?

nippoo:
The first point was fixed recently in https//github.com/kwikteam/phy/commit/09dfae4db06e7aa66e55e63e9881aa5721373d9e right? `pip install requirements-dev.txt` works for me now...

rossant:
oh, right

PR #518, opened 2015-07-07
[MRG] Moved flake8 options into setup.cfg

nippoo:


nippoo:
This closes #462 - development testing doesn't need `make` any more (`python setup.py flake8 && python setup.py test` will do)

PR #521, opened 2015-07-08
added nn probe files

neuromusic:
these are the probes in the nn database labeled "linear". I think some are actually custom, however, and are not accurate. further, I couldn't parse the logic for tetrodes or poly-trodes.


nippoo:
This looks awesome @neuromusic - thanks!

Where do I find a description of the probes? Or are these all the model names of each type? (i.e. will NN users know what they're looking for?)

rossant:
wow, that's a *lot* of probes!! A few remarks:

* Given the number of probes I think it would make sense to put them in a separate repo, to make it easier to change/add probes without making a new release. (btw we might also change the prb format someday...) Built-in utility functions would automatically download the requested probes (like we already do with sample data).

* I wonder if we need that many probes? We need two pieces of informations: (1) the graph and (2) the positions in any coordinate system. Only (1) is used by the spike detection algorithm; (2) is only used for visualization purposes (so it's not really critical). I imagine that all the 1x16, or all the 4x8, etc are all identical in this respect? Would it make sense to have dynamic probe generation utilities instead? Something like:

```
probe = multi_linear(n_shanks=4, n_channels_per_shank=8)
```

There could be support for common geometries, plus various utilities to create complex probes out of simpler components...

neuromusic:
these are the model names of each.

I'd actually argue that if you wrap in the delaunay tesselation, then only the geometry/coordinates are necessary, as the graph can then be computed. 

agreed that this is not the most efficient way to do this. the script I wrote to generate these simply took advantage of the "numshanks" and "sitespershank" parameters in the niPOD database.

I *do* like the idea of having utilities for constructing typical probe geometries, as you suggest. I also think it wouldn't be a bad idea to have some kind of a lookup so users can just call

``` python
probe = get_probe_from_name('A1x32-Poly3-6mm-60-177')
```
this is basically what I've done here...
https://github.com/gentnerlab/klusta-pipeline/blob/master/klusta_pipeline/probe.py#L13

neuromusic:
e.g.

``` python

probe_info = {
    'A4x8-5mm-50': {
        'type': 'linear',
        'n_shanks': 4,
        'n_sites_per': 8,
        'spacing': 50,
    },
    ...
}

get_probe(probe_name):
    info = probe_info[probe_name]
    if info['type']=='linear':
        return multi_linear(**info)
    else:
        raise UnknownTypeError

probe = get_probe('A4x8-5mm-50')
```

rossant:
your suggestions look good! 
cc @nsteinme

rossant:
closing in favor of #536

PR #539, opened 2015-07-15
[MRG] Refactor tempdir in testing suite

rossant:
Use pytest's fixture mechanism with a global `tempdir` fixture defined in `conftest.py`.


rossant:
If the CIs are happy, can we merge this quickly @nippoo ? I'll need it before I do further refactoring

PR #551, opened 2015-07-16
[MRG] Fixed bug in phy script

rossant:
The integration test has been removed -- it needs to be rewritten (generate a PRM instead of a KWIK).


rossant:
ready to merge @nippoo

PR #554, opened 2015-07-18
[MRG] Improve Qt testing suite

rossant:
* Use [pytest-qt](http://pytest-qt.readthedocs.org/) plugin: `qtbot` fixture in a test to register a widget, and emulate user actions
* No more hack with Qt tests
* Faster Qt tests
* Exceptions raised during a Qt test make the unit test fail: no more silent Qt fails
* Possibility to use a Qt bot: emulate keystrokes and mouse actions on any Qt view as part of the testing suite
* Fixed bug with snippets on Python 2
* Fixed unicode special character bug with cursor on Python 2 (the cursor is hidden on Python 2)


rossant:
ready to merge @nippoo

PR #559, opened 2015-07-22
[MRG] Do not crash if the .dat file is not present

### LINKED ISSUE #557: Exception in KwikModel when the dat file is not there

rossant:
Closes #557


rossant:
ready to merge @nippoo

PR #577, opened 2015-07-27
Fixed bug when using --interval that doesn't start at zero

### LINKED ISSUE #565: sampling an interval doesn't work if you don't start at zero
Line 97 of detect/spikedetekt.py is a bug: it checks that start is less than n_samples, but if an interval is used, n_samples is the number of samples within the interval whereas start is in absolute samples within the whole file. E.g. if you ask for the interval 9,12 with sample_rate=20000, then start will be 180000 but nsamples will only be 60000, so the assertion will fail though the data and numbers are fine. 

Also can we clear the NotImplementedError, line 100? I would like to use this function and I don't much care whether the spike times are referenced to the start of the interval or to the start of the total file (in fact maybe the former is preferable?).

nsteinme:
closes #565

### DIFF
--- a/phy/detect/spikedetekt.py
+++ b/phy/detect/spikedetekt.py
@@ -95,10 +95,10 @@ def _cut_traces(traces, interval_samples):
     else:
         start, end = 0, n_samples
     assert 0 <= start < end
-    # if start > 0:
+    if start > 0:
         # TODO: add offset to the spike samples...
-        # raise NotImplementedError("Need to add `start` to the "
-        #                          "spike samples")
+        raise NotImplementedError("Need to add `start` to the "
+                                 "spike samples")
     return traces, start
 
 
--- a/phy/detect/spikedetekt.py
+++ b/phy/detect/spikedetekt.py
@@ -98,7 +98,7 @@ def _cut_traces(traces, interval_samples):
     if start > 0:
         # TODO: add offset to the spike samples...
         raise NotImplementedError("Need to add `start` to the "
-                                 "spike samples")
+                                  "spike samples")
     return traces, start
 
 
--- a/phy/detect/spikedetekt.py
+++ b/phy/detect/spikedetekt.py
@@ -94,11 +94,11 @@ def _cut_traces(traces, interval_samples):
         n_samples = traces.shape[0]
     else:
         start, end = 0, n_samples
-    assert 0 <= start < n_samples
-    if start > 0:
+    assert 0 <= start < end
+    # if start > 0:
         # TODO: add offset to the spike samples...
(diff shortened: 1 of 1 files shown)


nsteinme:
Let me know if I created this properly... thanks!

nsteinme:
Hm, I put that NotImplementedError back in there - it really doesn't work without adding the offset to the spiketimes, because when you cluster-manual, it looks for the waveforms in the wrong place (or at least I assume that's what the error is).

rossant:
You can see that travis has failed; if you click on the red cross, you'll see that it's a lint error. The space that was removed line 100 should be put back, so that that second line is aligned to the line above.

On your local computer, you can run the tests with `make test-quick`, `make test`, or `python setup.py test`. If you just want to check the lint, do `make lint`.

rossant:
also, you can put in the first comment `closes #565`; then, when this PR is merged, that issue will be closed automatically.

otherwise the PR is good! You correctly used a new branch in your fork. Just a tiny comment: the branch is named `origin/fix-interval`, I guess some git command was not exactly right @nippoo ?

nsteinme:
K, that should work now. 

I used:
>> git checkout upstream/master
>> checkout -b origin/fix-interval

I put the "origin" there because I didn't want it to try to make the new branch in upstream; I wanted it to make it in origin instead. Which it seems to have done? I dunno.

rossant:
here's what I do generally:

```
git checkout master  # make sure I'm on local master
git pull upstream master  # synchronize master with upstream
git checkout -b mynewbranch  # create a new branch off master, no need to specify origin or upstream here because it's all local for now
# hack hack hack
git commit...
git push origin  # push the local branch to a new branch on origin (= your fork), it will create a new branch automatically
```

rossant:
ok, travis is still failing, but I think it's my fault :( for some reason master is broken, whereas all the PRs were passing (???). I've pushed a bug fix PR, if it works, I guess you'll have to rebase... I'll tell you how to do it (maybe connect on gitter)

nsteinme:
It will be easier for me to do a new branch with the one line fix ;)

rossant:
well actually I'm not sure about that: it should be as simple as `git fetch upstream` and `git rebase upstream/master`...

rossant:
@nsteinme  the fix PR has been merged, you can now `git fetch upstream` and `git rebase upstream/master`.

nsteinme:
K, done. Thanks -

Nick

rossant:
great, thanks! the next step is to `git push --force origin` to push your rebased branch on your fork. The `--force` is necessary because the history has been rewritten with the rebase

PR #580, opened 2015-07-27
[MRG] Bug fixes

rossant:


rossant:
actually merging now because master is currently broken! this should fix it...

PR #597, opened 2015-07-31
[MRG] Fix wizard when moving closest match cluster to group

### LINKED ISSUE #587: behavior incorrect when labeling "closest match" as MUA
As in KV, it should simply present you with the next closest match for the same best unsorted cluster. Instead, it picks a new best unsorted and presents you with that and its closest match. To get back to the previous one you must: backspace, shift+space, enter.

rossant:
Closes #587


rossant:
ready for review @nippoo

Issue #437, opened 2015-06-27
Better handling of raw data


rossant:
```
is there a path specified in the recordings[n]/raw/hdf5_path?
does it exist?
return(HDF5 file attribute)
is there a .dat or .bin of the same basename in the same path? (not sure how we deal with multiple recordings but k
return(.dat/bin)
is there a .raw.kwd of the same basename in the same path?
return(raw.kwd)
basically priority 1) hdf5 link 2) .dat/bin 3) .raw.kwd
```

nippoo:
The logic (inc #488) should be:

- is there a path specified in the recordings[n]/raw/hdf5_path?
- does this path exist? __if so, return it__
- if not, does a file with the same name exist in the same directory as the kwik file? __if so, return it___
- is there a .dat or .bin of the same basename in the same path? __if so, return it (maybe we will remove this because of complexities with multiple .dats!__
- is there a .raw.kwd of the same basename in the same path? __if so, return it__

PR #604, opened 2015-07-31
[MRG] Better handling of traces

### LINKED ISSUE #488: Fallback from absolute to relative raw data path in KwikModel

### LINKED ISSUE #590: Give warning and hide waveforms when KWD is missing
If the KWD is missing, there should be a warning and the WaveformView should be all black rather than show 0-height lines...

### LINKED ISSUE #592: Associated raw.kwd file needs to have the same name as the .kwik file, this should not be the case.
According to this function:

https://github.com/kwikteam/phy/blob/aaaa7f0e1b9ace264334c858f878c7c55e4b621f/phy/io/kwik/creator.py#L278-L293

The waveforms should be obtainable from any .raw.kwd. form that you associate with the .kwik file. 
In practice this did not work - the WaveformView in phy showed flat waveforms and the TraceView showed nothing. 

I created a smaller .kwik file containing only a subset of the spikes of a larger parent .kwik file
using the function I wrote below which utilises many functions of the creator.py class:

https://github.com/kwikteam/global_superclustering/blob/master/global_code/parallel_global.py#L39-L69

### LINKED ISSUE #602: Should only be able to open .kwik files, no .raw.kwd or .kwx.
At the moment, any file is copied to a backup before being processed. This should be limited to a .kwik.

nippoo:
Closes #602 
Closes #592 
Closes #590 
Finally fixes #488 and #437 properly!

Traces-opening logic is now (as per docstring)
        
> The `.kwik` and `.kwx` must be in the same folder with the same basename.

> The files containing the traces (`.raw.kwd` or `.dat` / `.bin`) are determined according to the following logic:

> - Is there a path specified to a file which exists in [KWIK]/recordings/[X]/raw? If so, open it.
> - If this file does not exist, does a file exist with the same name in the current directory? If so, open it.
> - If such a file does not exist, or no filename is specified in the [KWIK], then is there a `raw.kwd` with the experiment basename in the current directory? If so, open it.
> - If not, return with a warning.

### DIFF
--- a/phy/io/kwik/model.py
+++ b/phy/io/kwik/model.py
@@ -570,8 +570,20 @@ class KwikModel(BaseModel):
     def open(self, kwik_path, channel_group=None, clustering=None):
         """Open a Kwik dataset.
 
-        The `.kwik`, `.kwx`, and `.raw.kwd` must be in the same folder with the
+        The `.kwik` and `.kwx` must be in the same folder with the
         same basename.
+
+        The files containing the traces (`.raw.kwd` or `.dat` / `.bin`) are
+        determined according to the following logic:
+
+        - Is there a path specified in to a file which exists in [KWIK]/recordings/[X]/raw?
+        If so, open it.
+        - If this file does not exist, does a file exist with the same name in the current
+        directory? If so, open it.
+        - If such a file does not exist, or no filename is specified in the [KWIK], then
+        is there a `.dat`, `.bin, or .`raw.kwd` with the experiment basename in the current
+        directory? If so, open it.
+        - If not, return with a warning.
 
         Notes
         -----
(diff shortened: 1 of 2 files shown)


nippoo:
This is good to merge if tests are happy, @rossant

phy/io/kwik/model.py
```diff
@@ -174,39 +174,84 @@ def _open_h5_if_exists(kwik_path, file_type, mode=None):
...
+                debug("Loading traces: {}"
+                      .format(rel_path))
+                traces.append(kwd.read('/recordings/{}/data'
+                              .format(recording)))
+                continue
+        elif dat_path is not None:
+            rel_path = op.basename(dat_path)
+            rel_path = op.join(op.dirname(op.realpath(kwik.filename)),
+                               rel_path)
+            print('rel_path', rel_path)
```
rossant: to remove

phy/io/kwik/model.py
```diff
@@ -1136,9 +1193,7 @@ def n_clusters(self):
     # -------------------------------------------------------------------------
 
     def close(self):
-        """Close the `.kwik`, `.kwx`, and `.raw.kwd` files if they are open."""
+        """Close the `.kwik` and `.kwx` files if they are open."""
         if self._kwx is not None:
             self._kwx.close()
-        if self._kwd is not None:
-            self._kwd.close()
```
rossant: the trace file(s) are not closed?

nippoo: hrmmmmmmmmmm. I guess you're right, they should be closed. and the .dat as well...

nippoo:
Think both your issues are fixed, @rossant

Issue #465, opened 2015-07-01
`phy subcommand file` fails on Windows


rossant:
doing `phy describe file.kwik` or `phy cluster-manual file.kwik` or anything with a file, fails on Windows (it shows the CLI help).
@nsteinme @nippoo can you reproduce this on Windows?

nsteinme:
No, that doesn't happen for me. I have always been using this on windows. 

[phy] C:\Users\Nick>phy describe N:/M150218_NS1LAV/20150528_all.kwik
Kwik file               N:\M150218_NS1LAV\20150528_all.kwik
Recordings              1
List of shanks          1*
Clusterings             main*, original
Channels                120
Spikes                  2240058
Clusters                636
Duration                4292s

rossant:
cannot reproduce anymore, closing for now

nippoo:
I got this error yesterday with Tim Harris. Turns out they were using system python accidentally (some environment variable was causing the script to fail etc etc). This was on stock Ubuntu 14.04.

So I think this can happen if the version of python is wrong - or maybe argparse? We should check for this...

rossant:
hmm, not sure what exactly to check, but this should definitely be investigated

rossant:
@nippoo is there a way to run a particular unit test for many versions of python? typically we could have a test for the argparse code, and test it against all versions of python 2.7.x and 3.4.x...

nippoo:
Uh - Travis?

rossant:
Can run just one test instead of the full suite?

nippoo:
We already test against 2.7 and 3.4, no?

Nothing should functionally change between minor versions, so we should assume code that runs on 2.7.0 also does on 2.7.1 etc...

rossant:
Well i think travis already tests this, so it might be a minor version pb
Or does it only happen on windows? Then we should test with appveyor?

carlilek:
I have a request from Tim Harris's lab to run this on our compute cluster, which runs Scientific Linux 6.x (redistribution of RHEL). It is our policy to install from source because of our non-standard python locations. I believe I have all of the dependencies installed, as per the Advanced Install page. 

I am seeing the error in this thread using Python 2.7.8. When I issue any command from the command line with phy (other than phy -v), I get usage back rather than any sort of error or any output. 

[carlilek@h01u32 ~]$ export PATH=/usr/local/python-2.7.8/bin:$PATH
[carlilek@h01u32 ~]$ phy -v
0.2.2.dev0
[...]

PR #632, opened 2015-09-03
[MRG] Fix issue #465

nippoo:
The main parser's defaults overwrite the subparser so `func` is None on some systems (why this doesn't happen everywhere, I have no idea).


rossant:
cannot be merged yet?

nippoo:
Yep, good for merge

PR #646, opened 2015-09-16
Unified store

rossant:


rossant:
ready to merge @nippoo

phy/io/store.py
```diff
@@ -56,6 +59,65 @@ def _assert_per_cluster_data_compatible(d_0, d_1):
...
+            return
+        # Multiple arrays:
+        if self._contains_multiple_arrays(path):
+            return _load_arrays(path)
+        else:
+            return np.load(path)
+
+    def _delete_file(self, filename):
+        path = self._location(filename)
+        if op.exists(path):
```
nippoo: would be cleaner to do a try/except here? - what if it exists but permissions don't allow it to be deleted, or the delete fails for some reason or other

phy/io/store.py
```diff
@@ -56,6 +59,65 @@ def _assert_per_cluster_data_compatible(d_0, d_1):
...
+        if not op.exists(path):
+            debug("File `{}` doesn't exist.".format(path))
+            return
+        # Multiple arrays:
+        if self._contains_multiple_arrays(path):
+            return _load_arrays(path)
+        else:
+            return np.load(path)
+
+    def _delete_file(self, filename):
```
nippoo: might be worth commenting exactly what this function does:

- you cannot pass it a full pathname, just a file
- it will also delete .offsets

phy/io/store.py
rossant: yeah - actually I'll do that as part of the upcoming Refactoring...

PR #670, opened 2015-10-29
Fix #669

### LINKED ISSUE #669: Null adjacency graph results in no temporal floodfill

nippoo:
This manually adds trivial connectivity graphs into the FloodFillDetector by passing a `channels` attribute into the `connected_components()` method,  fixing issue #669 where having a null adjacency graph, or leaving a channel out, results in no temporal floodfill being carried out.

### DIFF
--- a/phy/traces/detect.py
+++ b/phy/traces/detect.py
@@ -155,7 +155,8 @@ def _to_list(x):
 def connected_components(weak_crossings=None,
                          strong_crossings=None,
                          probe_adjacency_list=None,
-                         join_size=None):
+                         join_size=None,
+                         channels=None):
     """Find all connected components in binary arrays of threshold crossings.
 
     Parameters
@@ -167,9 +168,12 @@ def connected_components(weak_crossings=None,
         `(n_samples, n_channels)` array with strong threshold crossings
     probe_adjacency_list : dict
         A dict `{channel: [neighbors]}`
+    channels : array
+        An (n_channels,) array with a list of all non-dead channels
     join_size : int
         The number of samples defining the tolerance in time for
         finding connected components
+
 
     Returns
     -------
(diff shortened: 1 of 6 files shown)


rossant:
travis is broken because of some setuptools bullsh*t -- see what I did at https://github.com/kwikteam/phy/commit/1b0d145965b173564cc2f67a64ed609c753312c7

rossant:
also could you please add a test case for non regression?

phy/traces/detect.py
```diff
@@ -190,6 +194,16 @@ def connected_components(weak_crossings=None,
...
+    if channels is None:
+        channels = []
+
+    # If the channels aren't referenced at all but exist in 'channels', add a
+    # trivial self-connection so temporal floodfill will work. If this channel
+    # is dead, it should be removed from 'channels'
+    for i in channels:
+        if not probe_adjacency_list.get(i):
+            probe_adjacency_list[i] = {i}
+
```
rossant: `probe_adjacency_list = {i for i in channels if not probe_adjacency_list.get(i, None)}` ?

nippoo:
thoughts on a good test for non-regression? in theory if the `mgraph` is non-null then we're all good so would a simple `assert mgraph` suffice?

the actual result of the floodfill failing is a bit more involved to test (you need to simulate data with two simultaneous strong threshold crossings and check that the connected components are merged correctly, but maybe this could also serve as a test for `join_size`)

can I just backport your travis fix?

rossant:
I think it is worth making a very minimal toy failing test case fixed by this PR, so that we're sure that future versions of the code don't have the same code (notably upcoming phy 1.0). Look at the unit tests of flood fill, there are many toy examples.

nippoo:
finally ready to merge @rossant

rossant:
Thanks! I shall not forget to forward-port this fix to the other branch.

PR #700, opened 2016-06-11
Faster memcache that requires hashable arguments

rossant:
memcache is now faster, but it requires function arguments to be hashable (**no ndarrays**). Typically, this is used for `cluster_id => value` functions so this is okay.

There should be performance improvements when a memcached function is called many times with different arguments (example of the similarity function). Previously, the bottleneck was the computation of the hash of the function's arguments (using joblib's `hash()` function, which is overkill unless using ndarrays)

### DIFF
--- a/phy/io/context.py
+++ b/phy/io/context.py
@@ -93,22 +93,19 @@ class Context(object):
                 dump(cache, fd)
 
     def memcache(self, f):
-        from joblib import hash
         name = _fullname(f)
         cache = self.load_memcache(name)
 
         @wraps(f)
-        def memcached(*args, **kwargs):
+        def memcached(*args):
             """Cache the function in memory."""
-            h = hash((args, kwargs))
-            if h in cache:
-                # logger.debug("Get %s(%s) from memcache.", name, str(args))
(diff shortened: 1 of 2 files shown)


nippoo:
This looks great! Do we never need to memcache ndarrays? I suppose we could always have a fallback or implement an alternative memcache-hash function if needed... I'm not at home atm but I'll check back later. Thanks!

rossant:
I don't think we need to memcache ndarrays. The other caching mechanism implemented uses joblib's memoization and it does support ndarrays. However it is much slower because it is more generic, and also it doesn't cache things in memory.

The similarity should be faster with this. 0.3 ms to load a memcached result seems slow. The computation of the args hash might have been the bottleneck.

nippoo:
OK - can confirm this has made the operation 60x quicker. Total time for similarity() is now 5-10ms rather than 300ms. Are you happy for me to merge this and https://github.com/kwikteam/phy-contrib/pull/46? Nice work!

rossant:
Terrific! yes feel free to merge this and kwikteam/phy-contrib#46

PR #701, opened 2016-07-01
Optimizations with large datasets

### LINKED ISSUE #42: Features improperly scaled?
In the feature view of `phy kwik-gui`, the scatter plots of selected clusters seems to have a different scale than the background noise. If I zoom out a lot (ctrl -) I can see clouds of point out of the noise, but even if I select all my clusters, I cannot find any unit corresponding to these point. All the selected features appear to have a different scale. A picture is probably better than any explanation, here  is a [screenshot](https://drive.google.com/file/d/0B0_6cZoUlMfQUGRMS2ZIUHJLOVE/view?usp=sharing).

If I zoom in a lot, it seems that the distribution of points is similar to that seen in the background, but at a different scale ([screenshot](https://drive.google.com/open?id=0B0_6cZoUlMfQWVEySnRkMzJmeWs)).

### LINKED ISSUE #43: in feature grid view, (0,0) is not at the center of the plot.
It appears to be somewhere below and to the left of the center of the plot, where "center of the plot" means the location of the crosshairs.

rossant:
## Optimizations
- Sparse waveforms
- Vectorize waveform filtering
- Cache CCGs
- In views, avoid unnecessary combine-split round-trip by handling cluster data separately
## Bug fixes
- Fixes https://github.com/kwikteam/phy-contrib/issues/43
- Fixes https://github.com/kwikteam/phy-contrib/issues/42


nsteinme:
Tested this with matching PR in phy-contrib. Looks great, works.

nsteinme:
Hey Cyrille, can we think about bringing this into the master? So other PRs can work with it? This one is essential for me. 

I think the precaching and stuff is great and there's no drawback to that for anyone, but I guess the only-show-16-channels part should be optional. What if there is a line you add in phy_config.py that says something like:

num_local_waveform_chans_to_display = 16

in which case it has the behavior that this PR currently has, or if that line is missing then it has the "usual" behavior which is to show samples that include all channels. 

The other thing is that I think we should go back to the waveform view styling from before: color channels according to the mask (as you calculate it in master is fine) and have the template part be split from the waveform part (switch between them with 'w'). 

What do you think?

rossant:
would you mind if instead of bringing back the w shortcut, we add a new TemplateView that shows just the templates? That would make the code simpler...

nsteinme:
Hm, I can see two problems than that. One, if you're short on monitor space
(common problem with phy since there's so much info to see) then you might
be annoyed about that, since you'd go from needing space for just one panel
to space for two, assuming you did want to see both. Two, there are some
times when you want to see how the templates relate to the extracted
waveforms, and being able to flip between the two quickly in the same view
allows that.

rossant:
That's fine I found a clean way to bring back that option.

rossant:
By default, all waveforms are loaded. If you want to only load unmasked channels, add this to your `params.py`:

```
max_n_unmasked_channels = 16  # maximum number of unmasked channels
waveform_mask_threshold = 0.1  # use None or 0 to load all waveforms, or a non-zero value to only load unmasked waveforms
```

rossant:
@nsteinme could you please update and try this to see if everything works fine? If so, I'll merge this PR.

nsteinme:
Hi Cyrille, it looks like it is loading all waveforms rather than the specified number: 
image


Maybe this is an interaction with the precaching plugin? My phy_config now has:
c.TemplateGUI.plugins = ['PrecachePlugin']

max_n_unmasked_channels = 16
waveform_mask_threshold = 0.1

rossant:
Can you delete `.phy` in your data dir?

nsteinme:
I did that, and it went through the whole precaching thing again taking
much longer than normal, then came up with this.

rossant:
I can't reproduce the problem, it works for me, even with the PreCache plugin...
Have you updated both phy and phycontrib?

nsteinme:
Yes, I did update both. I'm sure at least something did update because it
went from blanking out all channels outside the 16 to instead showing me
lots of waveform samples. But ok, I will try again from the top. Did I need
to do "python setup.py develop"? I didn't, but I'll try that.

rossant:
Maybe you should erase your branch completely and checkout it again from GitHub for both repos (I had to do it for some reason)

nsteinme:
Actually when I checked it out this time I had put it in a whole new branch already ("rossant-optim2"). Ok, but so I deleted both rossant-optim and rossant-optim2 branches, I pulled master to be current there, I ran python setup.py develop (all those steps on phy and phy-contrib). I deleted my .phy directory. I turned off the pre-caching plugin, and I changed max_n_unmasked_channels to 8 so it should be even easier to see whether it's working. But I still see the same thing: all channels' waveforms are loaded. And in fact, 16 channels are blue rather than the 8 that I asked for. Screenshot and also the text of my phy_config.py are below. What else to try? Where in the code did it specify the number 16 before - can we check whether the removal of that line has made it into my local copy and/or into this PR? Is it possible you had a typo in those variable names you suggested that I put into my phy_config? 

Thanks...

image

my phy_config.py:
```
# You can also put your plugins in ~/.phy/plugins/.

from phy import IPlugin

class MyPlugin(IPlugin):
    def attach_to_cli(self, cli):
        pass


c = get_config()
c.Plugins.dirs = [r'C:\Users\Nick/.phy/plugins/']

#c.TemplateGUI.plugins = ['PrecachePlugin']

max_n_unmasked_channels = 8  # maximum number of unmasked channels
waveform_mask_threshold = 0.2  # use None or 0 to load all waveforms, or a non-zero value to only load unmasked waveforms
```

rossant:
Are you using master? I realize I wasn't very clear: you should checkout again the optim branches in phy and phycontrib (I'll merge the branches to master as soon as things work for you!)

nsteinme:
Sorry, I didn't describe clearly what I did. It was: checkout master, pull, delete branch rossant-optim, delete branch rossant-optim2, make a new branch again called rossant-optim, pull from github.com/rossant/phy.git optim. You can see it all below. I did the same steps for both phy and phy-contrib. Probably I missed something...


```
C:\Users\Nick\Documents\GitHub\phy-contrib [rossant-optim2]> git checkout master

Switched to branch 'master'
Your branch is behind 'origin/master' by 24 commits, and can be fast-forwarded.
  (use "git pull" to update your local branch)
C:\Users\Nick\Documents\GitHub\phy-contrib [master]> git pull
Updating c2a858c..a72ec91
Fast-forward
 docs/index.md                                 |   6 +
 docs/screenshots/cluster_features.png         | Bin 0 -> 64690 bytes
 docs/screenshots/single_cluster.png           | Bin 0 -> 528111 bytes
 docs/screenshots/six_clusters.png             | Bin 0 -> 640379 bytes
 docs/screenshots/template_gui_traces_only.png | Bin 0 -> 232793 bytes
 docs/screenshots/two_clusters.png             | Bin 0 -> 550184 bytes
 docs/template-gui.md                          | 355 ++++++++++++++++++++++++++
 mkdocs.yml                                    |   5 +
 phycontrib/__init__.py                        |   2 +-
 phycontrib/kwik_gui/gui.py                    |   4 +-
 phycontrib/template/gui.py                    |   9 +
 phycontrib/utils/__init__.py                  |   1 +
 phycontrib/utils/saveprompt.py                |  25 ++
 setup.cfg                                     |   2 +-
 14 files changed, 406 insertions(+), 3 deletions(-)
 create mode 100644 docs/index.md
 create mode 100644 docs/screenshots/cluster_features.png
 create mode 100755 docs/screenshots/single_cluster.png
 create mode 100755 docs/screenshots/six_clusters.png
 create mode 100755 docs/screenshots/template_gui_traces_only.png
 create mode 100755 docs/screenshots/two_clusters.png
 create mode 100644 docs/template-gui.md
 create mode 100644 mkdocs.yml
 create mode 100644 phycontrib/utils/saveprompt.py
C:\Users\Nick\Documents\GitHub\phy-contrib [master]> git branch
* master
  rossant-optim
  rossant-optim2
C:\Users\Nick\Documents\GitHub\phy-contrib [master]> git branch -d rossant-optim

error: The branch 'rossant-optim' is not fully merged.
If you are sure you want to delete it, run 'git branch -D rossant-optim'.
C:\Users\Nick\Documents\GitHub\phy-contrib [master]> git branch -D rossant-optim

Deleted branch rossant-optim (was 5e4078a).
C:\Users\Nick\Documents\GitHub\phy-contrib [master]> git branch -D rossant-optim
2
Deleted branch rossant-optim2 (was b490f63).

C:\Users\Nick\Documents\GitHub\phy-contrib [master]> git checkout -b rossant-opt
im master
Switched to a new branch 'rossant-optim'
C:\Users\Nick\Documents\GitHub\phy-contrib [rossant-optim]> git pull https://git
hub.com/rossant/phy-contrib.git optim
From https://github.com/rossant/phy-contrib
 * branch            optim      -> FETCH_HEAD
Updating a72ec91..b490f63
Fast-forward
 phycontrib/kwik_gui/gui.py   |  33 ++++----
 phycontrib/template/gui.py   | 179 +++++++++++++++++++++++++++----------------
 phycontrib/utils/precache.py |   9 +++
 3 files changed, 142 insertions(+), 79 deletions(-)
```

rossant:
Can you check that your local copy of this file is up to date? https://github.com/kwikteam/phy-contrib/pull/49/files#diff-165c1daa21bc2c938ce5dc5f1b937318R393

Also could you do `import phycontrib; print(phycontrib.__file__)` and check that it's the right file?

nsteinme:
Yes, my local copy seems to have all the green segments, not the red ones,
at that link. Here's what I get:

In [1]: import phycontrib

In [2]: print(phycontrib.__file__)
c:\users\nick\documents\github\phy-contrib\phycontrib\__init__.py

which is the correct location.

rossant:
Oh I didn't realize you were putting the values in `phy_config`... you should put them directly in your `params.py` file...

nsteinme:
Aha! Sorry I didn't read that correctly. Maybe it is more of a phy_config type thing, in that it's a sort of global setting that one would like to use for every dataset? In any case, that's my fault, sorry!

Ok, so now the parameters are getting through and it is not loading all waveforms anymore, though I'm not sure they're behaving quite correctly. I set:
max_n_unmasked_channels = 8 
waveform_mask_threshold = 0.2

and I see:
image

The first question is, why are there 11 rather than 8? Second is, what is waveform_mask_threshold doing? It seems like it should have excluded channel 335 (and 333, 334, 346, 345) in favor of 340?

rossant:
I had a look at the code and here's what's happening. It's by no means ideal and I'm not sure how we came up with this complicated process (probably a lot of iterations that badly converged). I think this should be improved in another PR so if you agree we can merge this PR.

First, the exact value of `waveform_mask_threshold` is currently irrelevant. What's relevant is whether it's 0 or not (confusing).

For each template, I find the best channel (with the highest template amplitude). Then, I take the N nearest neighbors of this channel (depending on the 2D position on the probe). These are the unmasked channels for that template.

To each spike is assigned a static template (`spike_templates.npy`) which doesn't change during manual clustering).

When I select a cluster, I get a bunch of spikes who belong to one or several templates. For each spike, I get the masks associated to the associated template. Every channel that has at least one unmasked spike on it is selected for waveform loading. This is why you sometimes get more than 8 channels (because there are several templates with different unmasked channels associated to the cluster).

Feel free to create a new issue specifying a simpler and better way of selecting which channels you want to see in the waveform view for a given cluster.

nsteinme:
Ok, actually that all sounds quite reasonable to me! except the
waveform_mask_threshold thing (maybe just make it true/false?).

One thing is I do strongly recommend these parameters be globally set-able
rather than in params.py. I opened three different datasets yesterday and
each time had forgotten to change that particular dataset's params.py, and
each time had to go back to the previous dataset to load it's params.py to
remind myself exactly what the two variables were supposed to be called...

rossant:
But you mentioned earlier that it didn't seem to behave properly in your example?

OK for global params but I'll have to upgrade the config system first. In the meantime maybe I could change the defaults?

nsteinme:
Well I think the answer was: in the two constituent templates of the cluster in the last screenshot, the peak  channels must have been 341 and 337, neither of which had 340 in its closest 8, but which did have non-overlapping closest-8, so I got 11 total despite not having one that looked good. The solution is I needed to pick a number larger than 8 :)

PR #714, opened 2016-10-11
Improve manual clustering component

rossant:
*  Improve selection of next cluster after merge and move
* Add `label` action in `ManualClustering`
* Fixes kwikteam/phy-contrib#38
* Fixes kwikteam/phy-contrib#70

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,6 +18,7 @@ install:
   - conda update -q conda
   - conda info -a
   # Create the environment.
+  - conda install pyyaml
   - conda env create python=$TRAVIS_PYTHON_VERSION
   - source activate phy
   # Dev requirements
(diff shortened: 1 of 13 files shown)


rossant:
@nsteinme this is still experimental -- feel free to test

rossant:
@nsteinme in this PR, you can do `:l blah bleh` after selecting clusters, and it will create `cluster_blah.csv`. Feel free to try. You also need https://github.com/kwikteam/phy-contrib/pull/71

nsteinme:
Awesome, will try it!

nsteinme:
So this one is not working. In one situation it worked for me, but in other situations it selects the next cluster in _similarity view_ rather than the next in cluster view. I don't have a full theory of what works and what doesn't but if you try this I think it'll fail: select a cluster, spacebar, g [merge], spacebar, spacebar, alt+g. I think it will select a similarity view cluster rather than a cluster view cluster (it should always select a cluster view cluster!).

rossant:
ok, update and try again?

rossant:
@nsteinme `:label` now applies to clusters from the cluster view only

PR #722, opened 2016-10-31
Factor out and fix plotting of FeatureView labels

mspacek:
I've noticed two issues with the FeatureView labels, which make things very confusing for a new user:

1. Most of the labels are missing along the left edge of the subplots.
2. The label value in the bottom right corner is wrong.

This separates setting the FeatureView labels from the (iterated) plotting commands, gets rid of a hack, and fixes the two issues above. This is a rather quick and naive stab at the code, but it seems to work.

Here is what it looks like before the fix:

before

And here is what it looks like after:

after

### DIFF
--- a/phy/cluster/manual/views.py
+++ b/phy/cluster/manual/views.py
@@ -1099,21 +1099,17 @@ class FeatureView(ManualClusteringView):
 
     def _plot_labels(self, x_dim, y_dim):
         """Plot feature labels along left and bottom edge of subplots"""
-        dimlabels = []
-        for k in range(0, self.n_cols):
-            dimlabels.append(str(y_dim[(k, k)]))
-
         j = 0
         for i in range(0, self.n_cols):
             self[i, j].text(pos=[-1., 0.],
-                            text=dimlabels[i],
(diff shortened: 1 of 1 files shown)


rossant:
thanks! looks good, just a minor nitpick and it should be good to merge

mspacek:
I don't see offhand how those two loops could be merged, at least not without making the code harder to understand. It's one particular column and one particular row that we want to label, nothing more. The `dimlabels` thing was certainly unnecessary though, that's been removed.

rossant:
I just thought it would be a bit clearer to have a single loop over k with:
```python
self[k, 0].text()
self[self.n_cols - 1, k].text()
```

mspacek:
Right you are! OK, how's this?

rossant:
:+1: great! just make the linter happy and we're good :)

rossant:
in case you need it you can do `make lint` in the phy folder to check for lint errors locally

mspacek:
Thanks for the tip. Seems to have finally worked.

rossant:
thanks for your efforts!

PR #724, opened 2016-11-10
Refactor waveform loader

rossant:


rossant:
@nsteinme @marius10p @yger little question: how should I normalize waveforms (directly extracted from the raw data file) to compare them to templates? Can I assume that unwhitened templates and the traces are in the same space and can be directly compared, i.e. there's no need for a scaling factor? (for template subtraction in the trace view, and for displaying waveforms and templates in the same view)

yger:
In our case, but I guess this is the same for KiloSort, because we have non-binary amplitudes, there is, for every waveform, a scaling factor to establish a proper match with the template. To be more precise, when a waveform is detected at time step t with an amplitude k, it means that the spike in the raw data has been fitted by k.template. Most often, k is close to 1, but not exactly, because of drifts. So to be more precise, you should substract k*unwhitened template from the raw data if you want to have a precise view of what are the residuals considered by the algorithm in the TraceView. Conversly, when you are plotting groups of waveforms on top of the template in the Waveform view, then all those waveforms should be scaled by 1/k

Hope this helps

nsteinme:
Yes same for Kilosort. Those amplitude scaling factors - amplitudes.npy -
should in general be around 1 for Kilosort but I believe the mean of them
for any given template is not required to be.

Do you currently use a scaling factor?

yger:
What do you mean by "do we use" ? The answer is yes. SpyKING CIRCUS can
take care of small drifts or burst by using those scaling factors, and this
is a valuable information to see to understand what has been done by the
algorithm. In the MATLAB GUI, currently, waveforms are scaled by those
amplitudes values when compared to the template. Because for me cases
amplitudes can be pretty large, I strongly think they should be integrated
in the TraceView or the waveform view of phy. if not taken into account
when displaying the residuals, spikes can appear to still be there while in
fact they have been properly fitted. (We just submitted a small review
paper on biorxiv that should be online tomorrow I guess, discussing
template-matching based approaches and talk about those amplitudes. I can
send you the pdf.)

Best

nsteinme:
Sorry Pierre! I wasn't clear, I was addressing that question to Cyrille.
Because TraceView does already show the residual after subtracting the
templates, and it appears to do this [at least mostly] successfully so I
wondered whether Cyrille was already using a scaling factor. Definitely I
did not forget that Spyking Circus has always had this necessary feature :)

yger:
My bad :-) !  I am not sure that phy is handling amplitudes at the moment.
At least not in the waveform view, Cyrille, am I right ?

rossant:
I'm cleaning up the code a bit as it was developed organically, and I noticed that there is a scaling factor based on the mean waveform amplitude, the template amplitude, and other factors. I think it was mostly equivalent to the template amplitude, so I'll just use that now...

yger:
I'm not sure to understand what your scaling factor is.... This should be simpler, by just using the amplitudes plotted in the AmplitudeView

PR #735, opened 2017-01-04
[WIP] PyQt5 support

rossant:
Note: doesn't work yet on Ubuntu because of an [issue in Anaconda](https://github.com/ContinuumIO/anaconda-issues/issues/1267).


yger:
What about those that want to have Qt5 support without Anaconda?

2017-01-05 11:37 GMT+01:00 Cyrille Rossant <notifications@github.com>:

rossant:
What are the options to install PyQt5 without Anaconda on all platforms?

rossant:
The problem is that it is more work to support both PyQt4 and PyQt5 in the code. I would prefer to switch entirely to PyQt5 but only if everyone can use it...

yger:
No ok, I got your point. But in latest Ubuntu 16.10, PyQt5 is now there,
and in order to get phy running, I am forced to downgrade pyqt packages to
4 version. This is simply why I was asking. But no big deal, I can live
with this situation, and you can wait for a proper switch to Qt5 on all
platforms



2017-01-05 11:41 GMT+01:00 Cyrille Rossant <notifications@github.com>:

rossant:
It is a pain to be forced to use PyQt4, especially since Anaconda switched to PyQt5 by default (forced to downgrade packages) but at the same time PyQt5 is not fully supported on all platforms with OpenGL...

rossant:
Installing PyQt from pip instead of conda may actually work on Ubuntu. I'm making progress on the port, but it is not trivial. I'll see if it works on all platforms.

rossant:
This pull request has been merged into the `dev` branch.

Issue #813, opened 2019-02-06
Problem with setting bin and window


alejoe91:
Hi guys,

Probably (and hopefully) this is a quick one!

Using the dev version of phy and the template-gui phy-contrib (with pyqt5), when I try to set bin or set window fr the crosscorrelogram I get (using the `--debug` option):
`[W] actions:171          Invalid function arguments: expecting 1 but got 0`

Has anyone experienced the same issue? 

Thanks,
Alessio

rossant:
Thanks for the bug report, this should be fixed on the dev branch!

alejoe91:
Thanks! :)

PR #821, opened 2019-05-09
Fix bug in correlograms actions

alejoe91:
this fixes https://github.com/kwikteam/phy/issues/813 in the gloo branch

Without the `prompt=True` the window to change bin size and window size doesn't show

Cheers
Alessio


rossant:
thanks! (travis failure is unrelated, WIP)

PR #823, opened 2019-05-15
Calls to exists, load_json, save_json parse Path objects to str

alejoe91:
In Windows, `op.exists`, `_ensure_dir_exists`, `_load_json` and `save_json` can fail since `WindowsPath` are not handled by os. 

I just addeed a parse into `str` to the pathlib object when tose functions are called. This solves our problems in running phy on windows.

### DIFF
--- a/phy/gui/state.py
+++ b/phy/gui/state.py
@@ -43,7 +43,7 @@ def _gui_state_path(gui_name, config_dir=None):
 def _load_state(path):
     try:
         logger.debug("Load %s for GUIState.", path)
-        data = _load_json(path)
+        data = _load_json(str(path))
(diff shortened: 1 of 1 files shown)


alejoe91:
@rossant it doesn't seem that travis is failing because of this...or am I wrong?

rossant:
Thanks for the PR! Currently old-style `os.path` and new-style `pathlib` are mixed up in the code, I think I'll clean it properly

alejoe91:
Do you think you could merge this in the meanwhile so we can clone from this repo rather than our fork? :)

rossant:
sure no problem

PR #834, opened 2019-06-09
simpler installation process

m-beau:
Phy and phycontrib themselves are being installed through the environment file.

In the end, the use simply has to do: conda create -f path/to/environment.yml, and he is setup!

WARNING the @dev branches are used here, this will have to be changed.

### DIFF
--- a/environment.yml
+++ b/environment.yml
@@ -16,5 +16,4 @@ dependencies:
     - tqdm
     - joblib
     - click
-    - git+https://github.com/cortex-lab/phy.git@dev
-    - git+https://github.com/cortex-lab/phy-contrib.git@dev
+    - git+https://github.com/cortex-lab/phylib.git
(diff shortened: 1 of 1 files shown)


environment.yml
```diff
@@ -16,3 +16,5 @@ dependencies:
     - tqdm
     - joblib
     - click
+    - git+https://github.com/cortex-lab/phy.git@dev
+    - git+https://github.com/cortex-lab/phy-contrib.git@dev
```
rossant: to replace with `git+https://github.com/cortex-lab/phylib.git`
phy-contrib is deprecated in phy2

m-beau: Done

rossant: thanks but the first line (phy) was good!

rossant:
looks good!
the `.travis.yml` file should also be updated for continuous integration

- line 21: `- source activate phy2`
- remove line 26

I think that's all

rossant:
thanks! integrated in dev

PR #943, opened 2020-01-22
Cast positions to float32

alejoe91:


alejoe91:
https://github.com/cortex-lab/phylib/pull/19

PR #1167, opened 2022-07-18
Set cluster views from URL of local file

### LINKED ISSUE #1143: Blank ClusterView and SimilarityView
Hi,

I got some problem to start with a template-gui on my kilosort result. The ClusterView and SimilarityView didn't show anything. No mater how long I wait, it didn't change the blank content. If I tried to close the window, a timeout dialog pop up.

Screenshot_2022-03-11_19-43-55

Here are the part of packages installed in my conda environment.

```
# Name                    Version                   Build  Channel
pyqt5                     5.15.5                   pypi_0    pypi
pyqt5-qt                  5.15.2                   pypi_0    pypi
pyqt5-qt5                 5.15.2                   pypi_0    pypi
pyqt5-sip                 12.9.1                   pypi_0    pypi
pyqtwebengine             5.15.5                   pypi_0    pypi
pyqtwebengine-qt          5.15.2                   pypi_0    pypi
pyqtwebengine-qt5         5.15.2                   pypi_0    pypi
pytest-qt                 3.2.2                    pypi_0    pypi
qtconsole                 4.5.5                    pypi_0    pypi
phy                       2.0a1                    pypi_0    pypi
phylib                    2.4.1                    pypi_0    pypi
```

I guess it is something relate to the QtWebEngine which js wasn't run.

Thanks

### LINKED ISSUE #1151: cluster view and similarity view not shown
Hello

I am trying to use phy in the Mac mini with Big Sur as OS. I installed identical dependencies to the current Mac in another Mac, in which phy runs perfectly. However, in this Mac, all other function works except the cluster view and similarity view (screenshot attached). Your help will be greatly appreciated. Thank you.

Best,
phy_issue

### LINKED ISSUE #1156: ClusterView problem
Hello,
we have a similar problem like issue #1151 but with some interesting differences. We use a neuropixels probe and kilosort3. When we run phy on the output of kilosort , the ClusterView window remains blank, however, the SimilarityView window seems intact (see screenshot). Importantly, this issue only occurs in a part (~50%) of our recordings while other work nicely. When it occurs, it is independent of the machine we run phy at - currently we tested it at two Windows11 and two MacOS systems. When I compare the Kilosort outputs between the ‘working’ and ’not working’ measurements I also don’t see any obvious major differences. So far the mystery… Any suggestions or speculations?
Best!
Jan 
Screen Shot 2022-05-11 at 16 02 16 (2)

### LINKED ISSUE #1159: Cluster View on white when the file contains too many neurons
I'm implementing phyllum, a plugin on phy to automatically analyze neuropixel recordings in the cerebellum. This plugin computes for each cluster several metrics (more than 50), which are plotted on different columns of the cluster view. This plugin works perfectly for recordings in which the number of clusters is not too large (around 1400 or less). Nevertheless, if the file contains a lot of clusters, the cluster view only shows a white window. I have the same problem if a file contains a smaller number of neurons, but I try to generate more columns for each cluster. It seem like there is a limit in the maximum number of values that can be represented in this table: number of clusters x number of columns x number of values per column. I have verified (debugging the code) that the table containing all the information is properly generated, but the object that should plot the cluster view windows fails, making impossible to interact with phy. There is some way to increase the maximum number of values that this windows is able to represent?

Francisco.

### LINKED ISSUE #1165: ClusterView and SimilarityView views open entirely blank and unresponsive; issues exiting Phy window
I installed phy following the installation instructions with python 3.10.5 on Windows 10. I was attempting to use it with data produced by Kilosort, but ran into issues. I then tried to load the data from the old [kwikteam/phy-data](https://github.com/kwikteam/phy-data) repository, but ran into the same issue. The ClusterView and SimilarityView views were blank while loading up Phy; when using it with the data produced by kilosort, the FeatureView window did no load at all. Additionally, when attempting to exit the gui a number of exceptions popped up. I have included screenshots of the relevant screens, a zipped version of the data from kwikteam/phy-data, and the phy.log.

phy.log: https://gist.github.com/lukebemish/7dcb92a9b67ae6cb07de2ad097fac48
Screenshot (1)
unnamed
Screenshot 2022-07-01 151146
[template.zip](https://github.com/cortex-lab/phy/files/9031409/template.zip)

szapp:
This PR addresses and may resolve #1143, resolve #1151, resolve #1156, resolve #1159, and resolve #1165.

As @FranciscoNaveros and others have [observed](https://github.com/cortex-lab/phy/issues/1159#issue-1239068482), the issue of blank cluster views occurs when the number of clusters or columns exceeds a certain limit. This seems to be due to the method `setHtml` of `QWebEnginePage` having a [size limit of 2 MB](https://doc.qt.io/qtforpython-5/PySide2/QtWebEngineWidgets/QWebEnginePage.html?highlight=qwebenginepage#PySide2.QtWebEngineWidgets.PySide2.QtWebEngineWidgets.QWebEnginePage.setHtml) for the HTML code to load.

The proposed fix here simply stores the HTML code to a temporary file and uses the method `setUrl` instead. The temporary files are cleared appropriately (when new HTML is set and when phy closes with `_save_gui_state`).

This works reliably and solves the issue on a branch of an older version of phy I tested it on. I would appreciate if someone could confirm this on the latest version of phy before a merge is considered. There might also be a more elegant way to implement the temporary file handling.

### DIFF
--- a/phy/cluster/supervisor.py
+++ b/phy/cluster/supervisor.py
@@ -728,6 +728,10 @@ class Supervisor(object):
         """Save the GUI state with the cluster view and similarity view."""
         gui.state.update_view_state(self.cluster_view, self.cluster_view.state)
 
+        # Clear temporary HTML files
+        self.cluster_view.clear_temporary_files()
+        self.similarity_view.clear_temporary_files()
+
     def _get_similar_clusters(self, sender, cluster_id):
         """Return the clusters similar to a given cluster."""
         sim = self.similarity(cluster_id) or []
--- a/phy/gui/qt.py
+++ b/phy/gui/qt.py
@@ -13,7 +13,9 @@ import logging
 import os
 import os.path as op
 from pathlib import Path
+import shutil
 import sys
+import tempfile
 from timeit import default_timer
 import traceback
 
@@ -466,8 +468,21 @@ class WebView(QWebEngineView):
         self._callback = callback
         self.loadFinished.connect(self._loadFinished)
         static_dir = str(Path(__file__).parent / 'static') + '/'
-        base_url = QUrl().fromLocalFile(static_dir)
-        self.page().setHtml(html, base_url)
+
+        # Create local file from HTML
+        self.clear_temporary_files()
+        self._tempdir = Path(tempfile.mkdtemp())
+        shutil.copytree(static_dir, self._tempdir / 'html')
+        file_path = self._tempdir / 'html' / 'page.html'
+        with open(file_path, 'w') as f:
+            f.write(html)
+        file_url = QUrl().fromLocalFile(str(file_path))
+        self.page().setUrl(file_url)
+
+    def clear_temporary_files(self):
+        """Delete the temporary HTML files"""
+        if hasattr(self, '_tempdir') and self._tempdir.is_dir():
+            shutil.rmtree(self._tempdir, ignore_errors=True)
 
     def _callable(self, data):
         self.html = data


blinklab:
I have tested this solution with different large files that I could not open before and now they work perfectly.

Francisco Naveros.

elladulko:
Strangely we have one user where this fix appears to have worked but with an older user account CLusterView window and SimilarityView window are white despite both users using the same dataset in the same location (i.e., there is one dataset and not a copy of the dataset for the different users).

Antorithms:
In our case this does not solve the issue. We experience blank CLusterView that is "solved" by removing the metrics csv from the folder (while losing some unit information, e.g. ISI violations). How should we go about solving this ?
Thanks a lot!

Issue #1193, opened 2023-02-07
Attribute error on np.bool


Datamance:
Hi there,

I'm trying to use Phy in a virtual environment containing Python 3.11.1 and numpy 1.24.2. I get this error when I try to run the standard `phy template-gui params.py`

```
╰─(env) ○ phy template-gui params.py               
...
  File "/home/rico/.asdf/installs/python/3.11.1/lib/python3.11/site-packages/phy/plot/gloo/variable.py", line 86, in <module>
    gl.GL_BOOL: (1, gl.GL_BOOL, np.bool),
                                ^^^^^^^
  File "/home/rico/.asdf/installs/python/3.11.1/lib/python3.11/site-packages/numpy/__init__.py", line 305, in __getattr__
    raise AttributeError(__former_attrs__[attr])
AttributeError: module 'numpy' has no attribute 'bool'.

```

What is the maximum numpy version to which I must downgrade in order to get this working? Thanks.

Datamance:
Because of how the entrypoint generator is written, 

```python
  18   │ def importlib_load_entry_point(spec, group, name):
  19   │     dist_name, _, _ = spec.partition('==')
  20   │     matches = (
  21   │         entry_point
  22   │         for entry_point in distribution(dist_name).entry_points
  23   │         if entry_point.group == group and entry_point.name == name
  24   │     )
  25   │     return next(matches).load()
```

This is going to break every time. I'm not sure how this function is generated but perhaps you can tell setup.py to actually use the version information rather than throwing it away...

Datamance:
I'm not sure what's going on with how you load modules, but issues seem to arise with the use of virtual environments.

Even if I do `pip install phy --pre` without `--upgrade`, ensuring that I can pin numpy to a version that contains `np.bool` (I chose 1.22.4 to roughly match the date of the last commit in this repo), I still get this numpy compilation error:
```
RuntimeError: module compiled against API version 0x10 but this version of numpy is 0xf
```
upon running either `phy template-gui params.py` via the CLI or running it from a python script via
```python
from phy.apps.template import template_gui
template_gui("params.py")
```
To me, this suggests that `phy` is somehow trying to circumvent the `numpy` in the virtualenv and attempting to use the only updated one on my system - the one in my global python installation:
```
╭─rico at Five-Pebbles in ~/Science/Data/062122Ferret_Pen4_mseq1_g1/kilosort3 23-02-07 - 15:55:39
[...]

PR #1199, opened 2023-02-25
Update variable.py to. use np.bool_ rather than deprecated np.bool

zm711:
NumPy versions >1.24 will no longer work with code including np.bool rather than issue a warning. So I just changed np.bool to np.bool_ which should work for new and older versions of NumPy.


aepadilla4:
Hi @zm711! I don't have much experience with python. Would it be possible for you to share the steps on how to update the variable? I am having this same issue.

aepadilla4:
@zm711 Thank you so much!

Datamance:
#1193 Linking my issue so I know when this closes

zm711:
@carsen-stringer I don't know when there would be a new pypi release. There are a couple other features that are also missing from the current pip install (#1167 for example helps with users of linux). So if you have users that plan to use linux install from the `.git` might still be the better option until/if a new pypi release is made. (also as of this writing there is an issue with python 3.12)

PR #1210, opened 2023-06-15
updated environment.yml so people can also install env with that

zm711:
@rossant Welcome back!

I saw in the issues that some people were using the `environment.yml` file for install, so people don't use the yaml file with wrong versions I updated that.

In short I added all the dependencies you listed. Fixed the sklearn->scikit-learn and had it pull from git instead of pypi.

I just tested creating a conda env with 
`conda env create --file environment.yml`
and it worked fine on my computer.

### DIFF
--- a/environment.yml
+++ b/environment.yml
@@ -5,6 +5,7 @@ channels:
 dependencies:
   - python
   - pip
+  - git
   - numpy
   - matplotlib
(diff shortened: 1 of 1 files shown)


rossant:
thank you @zm711 !

PR #1229, opened 2023-11-02
Fix removal of imp in python 3.12

zm711:
@rossant,

I don't actually have the phy test suite setup so I made this fix based on reading python docs. In Python 3.12 they have completely removed `imp` and the rec is to switch to `importlib`. Could you either test to make sure these changes work or walk me through setting up the phy test suite so I can make sure this works. 

If you don't want to do this I'll do a PR to change the python version to < 3.11 for the install stuff.

### DIFF
--- a/phy/utils/plugin.py
+++ b/phy/utils/plugin.py
@@ -11,7 +11,7 @@ Code from http://eli.thegreenplace.net/2012/08/07/fundamental-concepts-of-plugin
 # Imports
 #------------------------------------------------------------------------------
 
-import imp
+import importlib
 import logging
 import os
 from pathlib import Path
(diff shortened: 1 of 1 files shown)


BisonLeo:
this patch tested working

zm711:
Thanks @BisonLeo , this code is called during plug-in usage. Do you actually use plug-ins to test that it works for plug-ins or did you test just to see if it would let you install with python 3.12?

AceTylerCholine:
I'm confused. For me, the rest of Phy doesn't work with Python 3.12, so I just define python=3.11 when installing it, but this patch prevents plugins from working for me on either 3.11 or 3.12

zm711:
Thanks @AceTylerCholine,

someone else said they tested it and it worked. Which is why it was merged. Could you open an issue an post the error so I can work on this some more. Could you also share your plugin so I have something to test against?

AceTylerCholine:
@zm711 https://github.com/cortex-lab/phy/issues/1281

PR #1230, opened 2023-11-02
Update requirements.txt

zm711:
Some people are following the developer instructions on the readme, which were missing a few packages in the requirements.txt, I've added the missing packages into the requirements, based on the conda environment.

I've also deleted the old install file inside of the `installer` folder, because some issues are coming up with people using the wrong `environment.yml` to install phy.


zm711:
I was still in the process of testing. Meant to have this as a draft for the moment.

zm711:
Okay I tested this locally and it install great as long as the version is <3.12.

zm711:
@rossant-- hope you're doing well. I think at least updating the requirements.txt will be beneficial. Currently I keep getting issue tracker problems where people try to pip install (even from source), but because the requirements.txt is not up-to-date, they get broken installs. I then have them do with the conda file and it works. So hopefully this will reduce some broken installs.

PR #1289, opened 2024-06-24
Add NumPy version limit

zm711:


zm711:
I need to add the numpy limit for now. Conda install works on linux but not windows. Not sure why yet.

PR #1298, opened 2024-08-02
Add initial Testing to Repo

zm711:


zm711:
Here is the log of the passing tests. I haven't been able to get all the tests working yet, but I figure something is better than nothing. 

https://github.com/zm711/phy/actions/runs/10220979583

PR #1354, opened 2025-10-17
Add codespell support (config, workflow to detect/not fix) and make it fix some typos

yarikoptic:
More about codespell: https://github.com/codespell-project/codespell .

I personally introduced it to dozens if not hundreds of projects already and so far only positive feedback.

CI workflow has 'permissions' set only to 'read' so also should be safe.

Overall summary with counts of what was fixed (ran before fixes introduced)

```
      1 attribure ==> attribute
      1 betwenn ==> between
      1 correspondance ==> correspondence
      1 detailled ==> detailed
      1 fot ==> for, fit, dot, rot, cot, got, tot, fog
      1 heigth ==> height
      1 inthe ==> in the
      1 lenght ==> length
      1 minimzing ==> minimizing
      1 optionnaly ==> optionally
      1 overlayed ==> overlaid
      1 recommanded ==> recommended
      1 shorcuts ==> shortcuts
      1 wich ==> which
      2 cacheing ==> caching
      2 currenty ==> currently
      2 discared ==> discarded, discard
      2 reduceable ==> reducible
      2 responsability ==> responsibility
      2 typicall ==> typically, typical
      3 specifiction ==> specification
      4 explicitely ==> explicitly
     86 overriden ==> overridden
```

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -3054,7 +3054,7 @@ Insert all GLSL snippets in a vertex and fragment shaders.
 
 **Notes**
 
-The vertex shader typicall contains `gl_Position = transform(data_var_name);`
+The vertex shader typically contains `gl_Position = transform(data_var_name);`
 which is automatically detected, and the GLSL transformations are inserted there.
 
 Snippets can contain `{{var}}` placeholders for the transformed variable name.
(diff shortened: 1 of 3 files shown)


zm711:
I support this in general, but I would still defer this to @rossant since I don't have enough time to support this repo currently.

PR #1401, opened 2026-07-28
Add an example recluster plugin for the Template GUI (ISO-SPLIT / Gaussian mixture)

Sachuriga:
## Summary

The Template GUI has no "recluster" action — the Kwik GUI's `shift+ctrl+K`, which
re-runs KlustaKwik2 on a selected cluster, has no counterpart. This adds one as an
**example plugin**: it re-clusters a selected cluster from the **PC features that
already exist** in the sorting output.

No spike detection is re-run, no raw data is read, and the result goes through the
normal `supervisor.actions.split`, so it is **undoable** like any manual split.

Split out of #1398 at the maintainer's request; #1398 now carries only the
packaging changes.

## What it does

- Default algorithm is **ISO-SPLIT** (the one MountainSort uses) when `isosplit`
  is installed. It is non-parametric — it decides the number of subclusters by
  testing each candidate split for unimodality — so it does not assume clusters
  are Gaussian. Bursting and drifting units are not, which is where a plain
  Gaussian mixture tends to over-split.
- A **Gaussian mixture** (BIC-selected) is kept as a fallback, and for the case
  where you want to impose the number of subclusters yourself.

## Keyboard shortcuts

| Shortcut       | Action          | Behaviour                                        |
| -------------- | --------------- | ------------------------------------------------ |
| `alt+k`        | `recluster`     | ISO-SPLIT, number of subclusters chosen automatically |
| `shift+alt+k`  | `recluster_n`   | Gaussian mixture into a number of subclusters you enter |
| `ctrl+alt+k`   | `recluster_dip` | ISO-SPLIT at a dip threshold you enter           |

## Dependencies

Optional, in a new `recluster` extra:

```toml
recluster = [
    "scikit-learn>=1.9.0",
    "isosplit==0.1.4",
]


rossant:
Thanks! I updated the PR branch on top of current master and added a few compatibility and safety guards plus one focused test. Could you review the edits and let us know what you think?

